PR #2, opened 2023-05-17
Update Clone.java

anna-geller:
adding an additional example, polishing the docs

### DIFF
--- a/src/main/java/io/kestra/plugin/git/Clone.java
+++ b/src/main/java/io/kestra/plugin/git/Clone.java
@@ -47,7 +47,7 @@ import javax.validation.constraints.NotNull;
                 "url: https://github.com/anna-geller/kestra-flows",
                 "branch: main",
                 "username: anna-geller",
-                "password: "{{ envs.personal_access_token }}""
+                "password: your_personal_access_token"
             }
         ),
         @Example(
--- a/src/main/java/io/kestra/plugin/git/Clone.java
+++ b/src/main/java/io/kestra/plugin/git/Clone.java
@@ -85,7 +85,6 @@ import javax.validation.constraints.NotNull;
                 "          - pandas"
             }
         )
-    }
 )
 public class Clone extends Task implements RunnableTask<Clone.Output> {
     @Schema(
--- a/src/main/java/io/kestra/plugin/git/Clone.java
+++ b/src/main/java/io/kestra/plugin/git/Clone.java
@@ -38,11 +38,10 @@ import javax.validation.constraints.NotNull;
             code = {
                 "url: https://github.com/dbt-labs/jaffle_shop",
                 "branch: main",
-                "username: dbt-labs"
             }
         ),
         @Example(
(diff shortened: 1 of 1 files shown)


src/main/java/io/kestra/plugin/git/Clone.java
```diff
@@ -98,25 +123,25 @@ public class Clone extends Task implements RunnableTask<Clone.Output> {
...
     @Schema(
-        title = "The username used to connect"
+        title = "The username or organization."
     )
     @PluginProperty(dynamic = true)
     private String username;
 
     @Schema(
-        title = "The password used to connect"
+        title = "The password or Personal Access Token."
```
Skraye, line 132: No need for uppercase here

src/main/java/io/kestra/plugin/git/Clone.java
```diff
@@ -34,16 +34,24 @@
...
-            title = "Clone a repository from http server",
+            title = "Clone a public GitHub repository",
             code = {
-                "url: https://github.com/kestra-io/plugin-template",
-                "branch: develop",
-                "username: <username>",
-                "password: <password>"
+                "url: https://github.com/dbt-labs/jaffle_shop",
+                "branch: main",
+                "username: dbt-labs"
```
Skraye, line 41: Useless username, bring confusion

src/main/java/io/kestra/plugin/git/Clone.java
```diff
@@ -34,16 +34,24 @@
...
             }
         ),
         @Example(
-            title = "Clone a repository from ssh server",
+            title = "Clone a private repository from an HTTP server such as a private GitHub repository using a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)",
```
Skraye, line 45: No need for uppercase here in `personal access token`

src/main/java/io/kestra/plugin/git/Clone.java
```diff
@@ -34,16 +34,24 @@
...
+            code = {
+                "url: https://github.com/anna-geller/kestra-flows",
+                "branch: main",
+                "username: anna-geller",
+                "password: your_personal_access_token"
+            }
+        ),
+        @Example(
+            title = "Clone a repository from an SSH server",
```
Skraye, line 54: Precise why you use directory

Skraye (changes requested):

src/main/java/io/kestra/plugin/git/Clone.java
anna-geller: good point, I thought this is required but just checked it works without it

src/main/java/io/kestra/plugin/git/Clone.java
anna-geller: wait I wasn't using a directory 😅 which line?

Skraye: I could have commented on a better line. Line 56, you use "directory: kestra", it will be better if, in your title, you explain what's his purpose.

anna-geller: added

PR #69, opened 2024-05-21
fix/add-revision-forecast

### LINKED ISSUE #68: Improve revision handling in the SyncFlows task
### Feature description

@brian-mulier-p I checked the `SyncFlows` implementation again and we need to find a better solution for `revision` handling. At the moment, we display state OVERWRITTEN and `revision 2` even though the synced flows haven't changed and have `revision 1`. This is not only confusing but also displaying `revision 2` while flows have only a single revision is incorrect/misleading. 

image

image

We should:
- either find a better solution to read each flow, see what changed, and only if it changed show state `OVERWRITTEN` and increment its revision; sure, some race condition could happen that while you check subsequent flows, the previous flow might have changed, but this is an edge case we can deliberately take out of scope for now
- or remove the revision property from the output (worst case, it would be better to have it; still, if we can't find a better revision handling, better to remove that from the output entirely)

Ofc it will take longer if we read the flow contents but optimizing the performance of this task is not an issue at this stage 👍

brian-mulier-p:
closes #68

### DIFF
--- a/src/main/java/io/kestra/plugin/git/AbstractSyncTask.java
+++ b/src/main/java/io/kestra/plugin/git/AbstractSyncTask.java
@@ -12,6 +12,7 @@ import lombok.Builder;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.experimental.SuperBuilder;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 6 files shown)


brian-mulier-p:
needs https://github.com/kestra-io/kestra/pull/3795 & https://github.com/kestra-io/kestra-ee/pull/1161

PR #82, opened 2024-06-21
fix(docs): correct examples to use company.team namespace

shrutimantri:


anna-geller:
This one is more complex, here we often push to dev and then sync to prod, I'll close this PR. thx for cross-checking together

PR #114, opened 2025-01-02
fix: rollback to tenantId() method

mgabelle:


loicmathieu (commented):
But WHY ?

mgabelle:
@loicmathieu I have the CI that is failing for an unknown reason. 
https://github.com/kestra-io/plugin-git/actions/runs/12579983597/job/35061119259
I wanted to check if it was the reason ;) 
Apparently not (I will close it)

PR #152, opened 2025-06-09
feat(git): add SyncFlow task for single flow sync

Malaydewangan09:
Adds a new task `io.kestra.plugin.git.SyncFlow` to allow syncing a single flow from a Git repository to a target Kestra namespace.

closes kestra-io/kestra-ee#3997

### DIFF
--- a/src/main/java/io/kestra/plugin/git/SyncFlow.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlow.java
@@ -1,0 +1,125 @@
+package io.kestra.plugin.git;
+
+import io.kestra.core.models.annotations.Example;
+import io.kestra.core.models.annotations.Plugin;
+import io.kestra.core.models.flows.Flow;
+import io.kestra.core.models.property.Property;
+import io.kestra.core.models.tasks.RunnableTask;
+import io.kestra.core.runners.DefaultRunContext;
+import io.kestra.core.runners.RunContext;
+import io.kestra.core.services.FlowService;
+import io.kestra.plugin.git.services.GitService;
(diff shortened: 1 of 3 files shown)


src/main/java/io/kestra/plugin/git/SyncFlow.java
```diff
@@ -0,0 +1,125 @@
...
+    private Property<String> flowPath;
+
+    @Schema(title = "If true, the task will only log the action without actually syncing the flow.")
+    @Builder.Default
+    private Property<Boolean> dryRun = Property.of(Boolean.FALSE);
+
+    @Override
+    public Output run(RunContext runContext) throws Exception {
+        GitService gitService = new GitService(this);
+        FlowService flowService = ((DefaultRunContext) runContext).getApplicationContext().getBean(FlowService.class);
```
fdelbrayelle, line 83: We started to discuss with @loicmathieu about avoiding to couple too strongly plugins with Kestra core so not very sure it's a good idea to use directly `FlowService` here.

src/test/java/io/kestra/plugin/git/SyncFlowTest.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Test
+    void createNewFlow() throws Exception {
+        RunContext runContext = runContext();
+
+        assertThat(flowRepositoryInterface.findAllForAllTenants().size(), is(0));
+
+        SyncFlow task = SyncFlow.builder()
+            .url(new Property<>("{{url}}"))
```
fdelbrayelle, line 48: `Property.ofExpression(...)` instead of `new Property<>(...)`

PR #185, opened 2025-09-04
chore(sanitycheck): add for git clone operation

Pradumnasaraf:
### What changes are being made and why?

Add sanaity check for `io.kestra.plugin.git.Clone`

### DIFF
--- a/src/test/resources/sanity-checks/clone.yml
+++ b/src/test/resources/sanity-checks/clone.yml
@@ -7,15 +7,19 @@ tasks:
     tasks:
       - id: clone
         type: io.kestra.plugin.git.Clone
-        url: https://github.com/dbt-labs/jaffle_shop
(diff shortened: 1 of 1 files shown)


Ben8t:
@Pradumnasaraf thanks! Can you add another assertion on at least a file please? To check that the clone actually worked as intenteded by adding file to the working dir, thanks!

Ben8t:
I see however the `exitCode == 0` doesn't test if the file is removed, only if the commands was successful. Subtle different I know, but believe me I've seen weird bugs here 😅
Best to add as simple check that a file exists explicitly

fdelbrayelle:
Thank you for this PR @Pradumnasaraf :wave: Feel free to assign @Malaydewangan09 and I to the PRs on plugins, storages and secrets repositories please :pray:

PR #251, opened 2026-02-24
fix(sync-namespace-files): avoid rewriting unchanged namespace files

fdelbrayelle:
fixes https://github.com/kestra-io/plugin-git/issues/248

### What changes are being made and why?

---

### How the changes have been QAed?

---

### Setup Instructions

---

### Contributor Checklist ✅

- [x] I have read and followed the [plugin contribution guidelines](https://kestra.io/docs/plugin-developer-guide/contribution-guidelines)

### DIFF
--- a/src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
+++ b/src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
@@ -6,8 +6,6 @@ import io.kestra.core.models.property.Property;
 import io.kestra.core.runners.RunContext;
 import io.kestra.core.storages.Namespace;
 import io.kestra.core.storages.NamespaceFile;
-import io.kestra.core.storages.StorageContext;
-import io.kestra.core.utils.WindowsUtils;
 import io.swagger.v3.oas.annotations.media.Schema;
 import lombok.*;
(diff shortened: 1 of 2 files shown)


src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
```diff
@@ -137,35 +138,79 @@ protected void deleteResource(RunContext runContext, String renderedNamespace, N
...
+    simulateResourceWrite(RunContext runContext, String renderedNamespace, URI uri, InputStream inputStream) throws IOException {
+        var namespace = runContext.storage().namespace(renderedNamespace);
+        var path = Path.of(uri.getPath());
+        var existingResource = this.findExistingResource(namespace, renderedNamespace, uri);
+
+        if (inputStream == null) {
+            return existingResource.orElseGet(() -> NamespaceFile.of(renderedNamespace, uri));
+        }
+
+        var gitContent = inputStream.readAllBytes();
```
proddata: reading the whole file in memory might be questionable, if the file size is not expected to be small (could be considered as source is git).

_Would probably be nice to use the git blob sha, however this would require change in core_

PR #312, opened 2026-07-23
fix: don't push draft flows to Git (#311)

### LINKED ISSUE #311: Ensure that Git Sync doesn't push Flows saved as Drafts to Git
We need to validate if this is happening and fix if drafts are pushed to git because they should not. We should only push to git flows that are fully saved, excluding drafts.

Malaydewangan09:
- closes #311


Malaydewangan09:
It will be merged when we bump this plugin to `2.0.0` core

fdelbrayelle (commented):
LGTM but:
- could you clearly document that draft flows aren't synced?
- check if it needs to be reported to plugin-ee-git?

PR #325, opened 2026-08-27
fix: fail fast with a clear error when no Kestra API authentication is available

fdelbrayelle:
## Why it failed

Since the SDK migration, `SyncFlows` (and other SDK-backed tasks: `SyncNamespaceFiles`, `SyncDashboards`, `Push*`, `NamespaceSync`) fail on Kestra 2.0 with an opaque `401 Unauthorized` when no Kestra API credentials are available. Reported via Slack: https://kestra-io.slack.com/archives/C04HTFM3VL6/p1787823820806449 (2.0.0-rc11).

Two causes:

1. `AbstractCloningTask.kestraClient()` silently returned an **unauthenticated** client when the `auth` property was null/empty and `runContext.sdk().defaultAuthentication()` found nothing — instead of failing with a clear configuration error, as the sibling `AbstractKestraTask` already does.
2. `kestra-api-client` 1.3.2: `KestraClientBuilder` defaults to `Auth.BASIC`, so the unauthenticated client sends `Authorization: Basic base64("null:null")` → the API answers 401 at call time, far from the actual misconfiguration.

## What was fixed

- `AbstractCloningTask.kestraClient()` now **fails fast** with `IllegalArgumentException` and an actionable message (set the task `auth` property — apiToken or username/password — or configure `kestra.tasks.sdk.authentication` defaults), matching `AbstractKestraTask` behavior.
- Guarded against `runContext.sdk()` being null (unwired contexts, e.g. tests): previously that path threw a `NullPointerException` before reaching the intended error.
- Bumped `kestraVersion` 2.0.0-rc10 → 2.0.0-rc11: the rc10 `worker`/`scheduler`/`executor` artifacts were never published to Maven Central, which broke test dependency resolution on CI (this also explains the red `main` branch builds).

Out of scope: the `kestra-api-client` default-to-`Auth.BASIC` bug should be fixed separately upstream (it should default to no auth header).

Companion PR: kestra-io/kestra#18746 adds `SDK.defaultAuthenticationOrThrow()` to core so every SDK-based plugin gets this fail-fast centrally. The fallback+throw here is consolidated into a single helper in `AbstractGitTask` (used by both `AbstractKestraTask` and `AbstractCloningTask`) and now delegates to it: `kestraVersion` is bumped to `2.0.0-SNAPSHOT` and the helper calls `SDK.defaultAuthenticationOrThrow()`. **CI on this PR stays red on compile until kestra-io/kestra#18746 is merged and a snapshot containing the method is published** — verified locally against a mavenLocal-published core from that branch. Merge order: kestra#18746 first, then re-run CI here.

## Test plan

- [x] `./gradlew compileJava compileTestJava` passes
- [x] New test `AbstractCloningTaskAuthTest#noAuthAndNoDefaultAuthentication_failsFast` (mirrors `AbstractKestraTaskAuthTest`): asserts `kestraClient()` throws `IllegalArgumentException` with the new message when no auth is set and no default authentication is configured — passes locally against rc11
- [x] Full CI suite green on rc11 (was unrunnable on rc10 — unresolvable test deps). Note: the fail-fast change required providing default SDK credentials in the test config, since many tests previously relied on the silently unauthenticated client against the mock API server.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 version=3.0.1-SNAPSHOT
-kestraVersion=2.0.0-rc10
+kestraVersion=2.0.0-rc11
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 version=3.0.1-SNAPSHOT
-kestraVersion=2.0.0-rc11
+kestraVersion=2.0.0-SNAPSHOT
--- a/src/main/java/io/kestra/plugin/git/AbstractCloningTask.java
+++ b/src/main/java/io/kestra/plugin/git/AbstractCloningTask.java
@@ -18,7 +18,6 @@ import io.kestra.core.exceptions.IllegalVariableEvaluationException;
 import io.kestra.core.models.annotations.PluginProperty;
 import io.kestra.core.models.property.Property;
 import io.kestra.core.runners.RunContext;
-import io.kestra.core.runners.SDK;
 import io.kestra.sdk.KestraClient;
 import io.kestra.sdk.internal.ApiException;
 import io.kestra.sdk.model.PagedResultsNamespace;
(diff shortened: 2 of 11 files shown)


Malaydewangan09:
Why not in core so all plugins, get this logic?

PR #333, opened 2026-09-04
fix(flows): avoid NPE in SyncFlows on repeated syncs

### LINKED ISSUE #327: Running multiple time the SyncFlows for same flow fails with NPE
### Describe the issue

Image

```
id: verify_sync_flows
namespace: verify.syncflows

tasks:
  - id: sync
    version: "3.0.0"
    url: https://github.com/brian-mulier-p/test-validation
    branch: master
    gitDirectory: flows
    targetNamespace: verify.syncflows.imported2
    dryRun: false
```

### Environment

- Kestra Version: develop

jymaire:
## Summary

- `SyncFlows` NPE'd on any run after the first: the "before" flow state is parsed from Kestra's export ZIP, which never carries a `revision`, and the sync-state decision unconditionally dereferenced `flowBeforeUpdate.getRevision()`.
- `wrapper()` now derives ADDED/UPDATED/UNCHANGED/DELETED from normalized source-text equality instead of revision arithmetic, so a null revision on either side can never throw.
- Replaced the ZIP-export-based single-flow lookup (`fetchSingleFlow`) with a direct call to the single-flow API endpoint (`flows().flow(...)`, same one used by `SyncFlow`), which returns an authoritative, non-null revision.
- `SyncResult.revision` is now documented as possibly absent when it can't be resolved.
- Added a `GET /flows/{namespace}/{id}` handler to the test `MockKestraApiServer` to back the new API call.
- Added regression tests: a unit test locking `wrapper()`'s null-revision/source-based behavior, container/mock tests running `SyncFlows` twice (dry-run and non-dry-run) against the same namespace, and a case where a flow is manually edited in Kestra between two runs.

closes: https://github.com/kestra-io/plugin-git/issues/327

## Test plan

- [x] `rtk test ./gradlew test` — compiles cleanly; the new `wrapper_withNullRevisionOnEitherSide_shouldNotThrowAndCompareBySource`, `shouldThrowOnInvalidFlowFromGit`, and `shouldIgnoreInvalidFlowFromGit` tests pass locally. The other `SyncFlowsTest`/`SyncFlowsContainerTest` cases (including the new "run twice" regression tests) require a real GitHub PAT (`GH_PERSONAL_TOKEN`) to clone the external `kestra-io/unit-tests` fixture repo and a live Kestra container — both unavailable in this sandbox, and this dependency is pre-existing (every git-network test in the repo fails identically here, before this change too). These run in CI via `secrets: inherit`.
- [ ] Manually verify: run `SyncFlows` against a real Kestra instance twice in a row on the same flow — second run must succeed and report `UNCHANGED`

---
*[View as Artifact](https://claude.ai/code/artifact/1ed9d84d-b37f-45f2-a46f-80b5a29f250f)*

### DIFF
--- a/src/main/java/io/kestra/plugin/git/SyncFlows.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlows.java
@@ -119,6 +119,7 @@ import lombok.experimental.SuperBuilder;
 )
 public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
     public static final Pattern NAMESPACE_FINDER_PATTERN = Pattern.compile("(?m)^namespace: (.*)$");
+    private static final Pattern REVISION_LINE_PATTERN = Pattern.compile("(?m)^revision:\\s*\\d+\\n?");
 
     @Schema(
         title = "Branch to sync",
@@ -208,7 +209,7 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         // Simulate: parse the source and determine if it differs from the current Kestra state
         var parsedFromSource = YamlParser.parse(flowSource, Flow.class);
         var tenantId = runContext.flowInfo().tenantId();
-        var currentFlow = fetchFlowFromApi(kestraClient(runContext), tenantId, parsedFromSource.getNamespace(), parsedFromSource.getId());
+        var currentFlow = fetchFlowFromApi(runContext, kestraClient(runContext), tenantId, parsedFromSource.getNamespace(), parsedFromSource.getId());
 
         if (currentFlow == null) {
             // New flow — revision will be 1 after import
@@ -234,7 +235,7 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         if (source == null) {
             return "";
         }
-        return source.replaceAll("(?m)^revision:\\s*\\d+\\n?", "").replace("\r\n", "\n").strip();
+        return REVISION_LINE_PATTERN.matcher(source).replaceAll("").replace("\r\n", "\n").strip();
     }
 
     @Override
@@ -282,7 +283,7 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         kestraClient.flows().importFlows(tenantId, true, toNamedTempFile(parsed.getId() + ".yaml", flowSource.stripTrailing()));
 
         // Re-fetch from Kestra to get the authoritative revision after import
-        var updated = fetchFlowFromApi(kestraClient, tenantId, parsed.getNamespace(), parsed.getId());
+        var updated = fetchFlowFromApi(runContext, kestraClient, tenantId, parsed.getNamespace(), parsed.getId());
         return updated != null ? updated : FlowWithSource.of(parsed, flowSource);
     }
 
@@ -384,7 +385,7 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         }
     }
 
-    private FlowWithSource fetchFlowFromApi(KestraClient kestraClient, String tenantId, String namespace, String flowId) {
+    private FlowWithSource fetchFlowFromApi(RunContext runContext, KestraClient kestraClient, String tenantId, String namespace, String flowId) {
         try {
             var apiFlow = kestraClient.flows().flow(namespace, flowId, tenantId, true, null, false);
             return FlowWithSource.builder()
@@ -395,7 +396,15 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
                 .source(apiFlow.getSource())
                 .build();
         } catch (ApiException e) {
-            // Flow does not exist yet, or could not be fetched
+            if (e.getCode() == 404) {
+                return null;
+            }
+            // Not a "does not exist" case: a permissions/API failure here would otherwise be silently
+            // misreported as the flow being ADDED, so surface it instead of swallowing it.
+            runContext.logger().warn(
+                "Failed to fetch flow {}.{} from the Kestra API (status {}): {}",
+                namespace, flowId, e.getCode(), e.getMessage()
+            );
             return null;
         }
     }
--- a/src/main/java/io/kestra/plugin/git/SyncFlows.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlows.java
@@ -209,9 +209,14 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         // Simulate: parse the source and determine if it differs from the current Kestra state
         var parsedFromSource = YamlParser.parse(flowSource, Flow.class);
         var tenantId = runContext.flowInfo().tenantId();
-        var currentFlow = fetchFlowFromApi(runContext, kestraClient(runContext), tenantId, parsedFromSource.getNamespace(), parsedFromSource.getId());
+        var lookup = fetchFlowFromApi(runContext, kestraClient(runContext), tenantId, parsedFromSource.getNamespace(), parsedFromSource.getId());
+        var currentFlow = lookup.flow();
 
         if (currentFlow == null) {
+            if (!lookup.resolved()) {
+                // Lookup failed for a reason other than "flow does not exist" — don't fabricate a revision
+                return FlowWithSource.of(parsedFromSource, flowSource);
+            }
             // New flow — revision will be 1 after import
             return simulatedFlow(parsedFromSource, flowSource, 1);
         }
@@ -225,6 +230,9 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         // Changed — simulate next revision
         int projectedRevision = currentFlow.getRevision() != null ? currentFlow.getRevision() + 1 : 1;
         return simulatedFlow(parsedFromSource, flowSource, projectedRevision);
+    }
+
+    private record FlowLookup(FlowWithSource flow, boolean resolved) {
     }
 
     private static FlowWithSource simulatedFlow(Flow parsedFromSource, String flowSource, int revision) {
@@ -283,7 +291,7 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         kestraClient.flows().importFlows(tenantId, true, toNamedTempFile(parsed.getId() + ".yaml", flowSource.stripTrailing()));
 
         // Re-fetch from Kestra to get the authoritative revision after import
-        var updated = fetchFlowFromApi(runContext, kestraClient, tenantId, parsed.getNamespace(), parsed.getId());
+        var updated = fetchFlowFromApi(runContext, kestraClient, tenantId, parsed.getNamespace(), parsed.getId()).flow();
         return updated != null ? updated : FlowWithSource.of(parsed, flowSource);
     }
 
@@ -385,27 +393,30 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         }
     }
 
-    private FlowWithSource fetchFlowFromApi(RunContext runContext, KestraClient kestraClient, String tenantId, String namespace, String flowId) {
+    private FlowLookup fetchFlowFromApi(RunContext runContext, KestraClient kestraClient, String tenantId, String namespace, String flowId) {
         try {
             var apiFlow = kestraClient.flows().flow(namespace, flowId, tenantId, true, null, false);
-            return FlowWithSource.builder()
-                .id(apiFlow.getId())
-                .namespace(apiFlow.getNamespace())
-                .revision(apiFlow.getRevision())
-                .tenantId(tenantId)
-                .source(apiFlow.getSource())
-                .build();
+            return new FlowLookup(
+                FlowWithSource.builder()
+                    .id(apiFlow.getId())
+                    .namespace(apiFlow.getNamespace())
+                    .revision(apiFlow.getRevision())
+                    .tenantId(tenantId)
+                    .source(apiFlow.getSource())
+                    .build(),
+                true
+            );
         } catch (ApiException e) {
             if (e.getCode() == 404) {
-                return null;
+                return new FlowLookup(null, true);
             }
-            // Only 404 means "flow does not exist" and is swallowed silently above; any other status is a
-            // real API/permissions failure, so log it and still return null (sync continues, flow treated as absent).
+            // Only 404 means "flow does not exist"; any other status is a real API/permissions failure that
+            // leaves the lookup unresolved, so callers must not assume the flow is absent.
             runContext.logger().warn(
                 "Failed to fetch flow {}.{} from the Kestra API (status {}): {}",
                 namespace, flowId, e.getCode(), e.getMessage()
             );
-            return null;
+            return new FlowLookup(null, false);
         }
     }
 
--- a/src/main/java/io/kestra/plugin/git/SyncFlows.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlows.java
@@ -232,9 +232,6 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         return simulatedFlow(parsedFromSource, flowSource, projectedRevision);
     }
 
-    private record FlowLookup(FlowWithSource flow, boolean resolved) {
-    }
-
     private static FlowWithSource simulatedFlow(Flow parsedFromSource, String flowSource, int revision) {
         return FlowWithSource.of(parsedFromSource, flowSource).toBuilder().revision(revision).build();
     }
@@ -465,6 +462,9 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         }
     }
 
+    private record FlowLookup(FlowWithSource flow, boolean resolved) {
+    }
+
     @SuperBuilder
     @Getter
     public static class SyncResult extends AbstractSyncTask.SyncResult {
@@ -473,7 +473,8 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
 
         @Schema(
             title = "Flow revision",
-            description = "Revision of the flow after sync. May be absent if it could not be resolved from the Kestra API."
+            description = """
+                Revision of the flow after sync. May be absent if it could not be resolved from the Kestra API."""
         )
         private Integer revision;
     }
--- a/src/main/java/io/kestra/plugin/git/SyncFlows.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlows.java
@@ -399,8 +399,8 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
             if (e.getCode() == 404) {
                 return null;
             }
-            // Not a "does not exist" case: a permissions/API failure here would otherwise be silently
-            // misreported as the flow being ADDED, so surface it instead of swallowing it.
+            // Only 404 means "flow does not exist" and is swallowed silently above; any other status is a
+            // real API/permissions failure, so log it and still return null (sync continues, flow treated as absent).
             runContext.logger().warn(
                 "Failed to fetch flow {}.{} from the Kestra API (status {}): {}",
                 namespace, flowId, e.getCode(), e.getMessage()
--- a/src/main/java/io/kestra/plugin/git/SyncFlows.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlows.java
@@ -208,25 +208,33 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         // Simulate: parse the source and determine if it differs from the current Kestra state
         var parsedFromSource = YamlParser.parse(flowSource, Flow.class);
         var tenantId = runContext.flowInfo().tenantId();
-        var currentFlow = fetchSingleFlow(kestraClient(runContext), tenantId, parsedFromSource.getNamespace(), parsedFromSource.getId());
+        var currentFlow = fetchFlowFromApi(kestraClient(runContext), tenantId, parsedFromSource.getNamespace(), parsedFromSource.getId());
 
         if (currentFlow == null) {
             // New flow — revision will be 1 after import
-            return parsedFromSource.toBuilder().revision(1).build();
+            return simulatedFlow(parsedFromSource, flowSource, 1);
         }
 
-        // Compare normalised source to detect changes
-        // Strip revision: from the exported source — the Kestra ZIP export injects it but git sources never have it
-        var normalised = flowSource.replace("\r\n", "\n").strip();
-        var currentSourceWithoutRevision = currentFlow.getSource().replaceAll("(?m)^revision:\\s*\\d+\\n?", "");
-        var currentNormalised = currentSourceWithoutRevision.replace("\r\n", "\n").strip();
-        if (normalised.equals(currentNormalised)) {
+        // Compare normalised source (some Kestra versions inject a leading revision: line into the fetched source; git sources never have it) to detect changes
+        if (normalizeSource(flowSource).equals(normalizeSource(currentFlow.getSource()))) {
             // Unchanged — return the current flow so wrapper detects UNCHANGED
             return currentFlow;
         }
 
         // Changed — simulate next revision
-        return parsedFromSource.toBuilder().revision(currentFlow.getRevision() + 1).build();
+        int projectedRevision = currentFlow.getRevision() != null ? currentFlow.getRevision() + 1 : 1;
+        return simulatedFlow(parsedFromSource, flowSource, projectedRevision);
+    }
+
+    private static FlowWithSource simulatedFlow(Flow parsedFromSource, String flowSource, int revision) {
+        return FlowWithSource.of(parsedFromSource, flowSource).toBuilder().revision(revision).build();
+    }
+
+    private static String normalizeSource(String source) {
+        if (source == null) {
+            return "";
+        }
+        return source.replaceAll("(?m)^revision:\\s*\\d+\\n?", "").replace("\r\n", "\n").strip();
     }
 
     @Override
@@ -270,13 +278,12 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
             throw new FlowProcessingException("Invalid flow imported from Git (" + ref + "): " + flowValidated.getConstraints());
         }
 
-        var parsedId = YamlParser.parse(flowSource, Flow.class).getId();
-        kestraClient.flows().importFlows(tenantId, true, toNamedTempFile(parsedId + ".yaml", flowSource.stripTrailing()));
+        var parsed = YamlParser.parse(flowSource, Flow.class);
+        kestraClient.flows().importFlows(tenantId, true, toNamedTempFile(parsed.getId() + ".yaml", flowSource.stripTrailing()));
 
-        // Re-fetch from Kestra to get the updated revision
-        var parsedNamespace = YamlParser.parse(flowSource, Flow.class).getNamespace();
-        var updated = fetchSingleFlow(kestraClient, tenantId, parsedNamespace, parsedId);
-        return updated != null ? updated : YamlParser.parse(flowSource, Flow.class);
+        // Re-fetch from Kestra to get the authoritative revision after import
+        var updated = fetchFlowFromApi(kestraClient, tenantId, parsed.getNamespace(), parsed.getId());
+        return updated != null ? updated : FlowWithSource.of(parsed, flowSource);
     }
 
     private static String replaceNamespace(String renderedNamespace, URI uri, InputStream inputStream) throws IOException {
@@ -303,7 +310,8 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
             syncState = SyncState.DELETED;
         } else if (flowBeforeUpdate == null) {
             syncState = SyncState.ADDED;
-        } else if (flowBeforeUpdate.getRevision().equals(Objects.requireNonNull(flowAfterUpdate).getRevision())) {
+        } else if (flowAfterUpdate != null && normalizeSource(flowBeforeUpdate.getSource()).equals(normalizeSource(flowAfterUpdate.getSource()))) {
+            // Source-based comparison: revision may be null when a flow is parsed from an export and cannot be used for diffing
             syncState = SyncState.UNCHANGED;
         } else {
             syncState = SyncState.UPDATED;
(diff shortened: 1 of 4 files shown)


jymaire:
## CI failure is unrelated to this diff

`check / main` is red with 2 failures, both dashboard container tests:

- `PushDashboardsContainerTest.pushDashboardsToGit_shouldProduceCommitUrl()`
- `SyncDashboardsContainerTest.syncDashboardsFromGit_shouldProduceDiffOutput()`

Both fail on the same server-side response:

```
405 Method [POST] not allowed for URI [/api/v1/main/dashboards]. Allowed methods: [HEAD, GET]
```

**Cause:** `AbstractKestraContainerTest` pins the floating tag `kestra/kestra:develop-slim`. That image was re-pushed at `2026-09-04T08:11:41Z`; this PR's CI run started at `08:20:26Z` and is the first run to pull it. The last green run on `main` (`33836053078`) started at `04:14:07Z`, before the push. The dashboards endpoint no longer accepts `POST` in that image.

**Why it is not this PR:** the diff touches `SyncFlows.java`, `MockKestraApiServer.java`, `SyncFlowsTest.java` and `SyncFlowsContainerTest.java` only. Nothing here can change a real Kestra server's HTTP routing, and `MockKestraApiServer` is not used by container tests at all.

**Evidence the fix itself is verified:** `main` runs 98 tests, all passing. This PR runs 103 — the 5 added tests all executed (0 skipped) and passed against a real Kestra container, including `syncFlowsTwice_shouldReportUnchangedOnSecondRun`, which is the direct reproduction of #327. The 2 dashboard failures are the only delta.

The dashboards breakage needs handling separately, not in this PR.

jymaire (commented):
## Kestra Plugin Code Review

*(Note: this review was posted with the COMMENT event due to a tooling issue while drafting; treat the verdict line below as authoritative.)*

The fix correctly addresses issue #327. `wrapper()`'s UNCHANGED/UPDATED decision is now derived from normalized source-text equality and never dereferences a nullable `getRevision()`. The second latent NPE on the dry-run path (`currentFlow.getRevision() + 1`) is also guarded (`currentFlow.getRevision() != null ? ... : 1`). The ZIP-export-based `fetchSingleFlow` was correctly replaced with `fetchFlowFromApi`, hitting the same single-flow endpoint already used by `SyncFlow`, giving an authoritative non-null revision.

Traced all four `wrapper()` cascade branches (ADDED / UPDATED / UNCHANGED / DELETED) plus the `delete: true` path — no remaining `getRevision()` null-dereference. Verified that the simulated/written "after" flow objects are built via `FlowWithSource.of(...)` (not a plain `Flow.toBuilder()...build()`, whose base `getSource()` always returns `null` per core's `Flow.java`), so the new source-comparison in `wrapper()` always has a real source to compare on both sides — this is a necessary companion fix and it's handled correctly in every code path (`simulatedFlow`, `writeResource`'s fallback, and the API-fetched `currentFlow`).

No N+1 regression: `fetchFlowFromApi` replaces one per-flow ZIP-export API call with one per-flow single-flow GET — same call cardinality, lighter payload, no new inefficiency.

The three new regression tests in `SyncFlowsTest` (`runTwice_shouldReportUnchangedOnSecondRun`, `dryRun_afterManualEditInKestra_shouldProjectIncrementedRevisionWithoutThrowing`, `wrapper_withNullRevisionOnEitherSide_shouldNotThrowAndCompareBySource`) each target a distinct NPE site that existed pre-fix and would genuinely throw with the old code — not tautological. Container-test mirrors follow the existing `kestra.git.pat` / `secrets: inherit` convention used by every other `*ContainerTest` in this plugin, so they run unconditionally in CI (not silently skipped behind an env gate).

Agree the 2 CI failures (`PushDashboardsContainerTest`, `SyncDashboardsContainerTest`) are unrelated — no dashboard code is touched by this diff, and the described 405 from a re-pushed `develop-slim` image is a plausible independent cause.

🟡 `SyncFlows.java:233-238` (`normalizeSource`) — `source.replaceAll("(?m)^revision:\\s*\\d+\\n?", "")` compiles a new `Pattern` on every call. It's now invoked on both sides of the comparison in `wrapper()` and in `simulateResourceWrite`/`writeResource`, so it runs multiple times per synced flow. Hoist to a `private static final Pattern REVISION_LINE = Pattern.compile(...)` per the "regex compiled per iteration" performance guideline. Low impact at current call volumes, non-blocking.

### Security (OWASP Top 10:2025 + KPS) — 1 finding
🟡 `SyncFlows.java:387-401` (`fetchFlowFromApi`) — `catch (ApiException e) { return null; }` treats every failure identically: a genuine 404 (flow not yet synced), a 401/403 (bad or expired token), and a 5xx from the Kestra API all collapse into "flow doesn't exist," with zero logging. On the dry-run path (`simulateResourceWrite`) this means a transient auth/API failure would silently report every already-existing flow as `ADDED` with revision 1 in the diff output, with no trace of why. This mirrors the pre-existing pattern in `SyncFlow.java:163`, so it's not a new class of bug introduced by this diff, but worth tightening: check `e.getCode() == 404` and only treat that as "not found," logging or rethrowing on any other status code. Non-blocking for this PR since it's consistent with established convention, but worth a follow-up covering both `SyncFlows` and `SyncFlow`.

### Performance — 0 additional findings
No unbounded allocation, no new blocking-in-loop pattern, no thread-safety concern introduced by this diff.

**Verdict:** APPROVE

jymaire (commented):
.

src/test/java/io/kestra/plugin/git/SyncFlowsTest.java
```diff
@@ -635,6 +643,226 @@ void shouldThrowOnInvalidFlowFromGit() throws Exception {
...
+        task.run(runContext());
+
+        server.forceGetFlowStatus(NAMESPACE, "unchanged-flow", 500);
+
+        SyncFlows dryRunTask = task.toBuilder().dryRun(Property.ofValue(true)).build();
+        RunContext dryRunContext = runContext();
+        SyncFlows.Output dryOutput = dryRunTask.run(dryRunContext);
+
+        // Behavior is preserved (no exception), but the failure must not be silently misreported without a trace
+        Map<String, Object> diff = findDiffByFlowId(dryRunContext, dryOutput.diffFileUri(), "unchanged-flow");
```
jymaire, line 774: 🟠 [Guidelines/Correctness] Assertion likely wrong — actual outcome should be UNCHANGED, not ADDED
Problem: wrapper() derives syncState from flowBeforeUpdate, which is populated in AbstractSyncTask.run() via fetchResources() (the ZIP-export path). forceGetFlowStatus() only intercepts the single-flow GET /flows/{ns}/{id} handler used inside fetchFlowFromApi — the export endpoint is routed separately in MockKestraApiServer and is never affected by it. Since the baseline run already imported "unchanged-flow" identically to the git fixture and nothing changed on either side before the forced-500 dry run, flowBeforeUpdate will be non-null with a source matching flowAfterUpdate's source, so wrapper()'s else-if chain can never reach ADDED (which requires flowBeforeUpdate == null) — it should land on UNCHANGED instead. This matches the PR description's own admission that this test has not actually been observed passing.
Fix: Run this locally with a real GH_PERSONAL_TOKEN (or re-trace it) before merge and correct the expected syncState (very likely UNCHANGED). If the intent is to also exercise the revision-projection defect, combine this scenario with manuallyEditFlowInKestra (like the sibling test) so the before/after sources genuinely differ, and assert the resulting (likely wrong) revision plus the warning log, rather than asserting ADDED.

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -376,30 +385,26 @@ private List<Flow> fetchFlowsFromKestra(KestraClient kestraClient, RunContext ru
...
+                .source(apiFlow.getSource())
+                .build();
+        } catch (ApiException e) {
+            if (e.getCode() == 404) {
+                return null;
             }
-            return null;
-        } catch (Exception e) {
+            // Not a "does not exist" case: a permissions/API failure here would otherwise be silently
+            // misreported as the flow being ADDED, so surface it instead of swallowing it.
```
jymaire: 🟡 [Guidelines] Comment overstates the actual blast radius of a non-404 failure
Problem: this comment claims a permissions/API failure would be misreported as the flow being ADDED. In practice wrapper() decides ADDED/UPDATED/UNCHANGED from the independently-fetched (ZIP export) before-state, not from this method's return value, so a non-404 failure here cannot flip an existing flow's diff to ADDED — it can only cause the projected revision (simulateResourceWrite's fallback-to-1, or writeResource's fallback FlowWithSource.of(parsed, flowSource) with no revision) to be wrong for an otherwise-correctly-classified UPDATED/UNCHANGED entry.
Fix: tighten the comment to describe the actual effect (a wrong/missing projected revision on a real API failure), so the next reader isn't misled about what this fix actually guards against.

jymaire (commented):
## Kestra Plugin Code Review — Cycle 2 (verifying cycle-1 follow-up fixes, commit `31bc53f`)

Both cycle-1 🟡 suggestions are addressed at the production-code level: `REVISION_LINE_PATTERN` is now a precompiled `static final Pattern` used via `.matcher(source)` (no `Matcher` hoisted to a field — thread-safe), and `fetchFlowFromApi` now narrows the swallow to `e.getCode() == 404` only, logging a `WARN` with namespace/flowId/status for anything else, threaded correctly through both call sites (`simulateResourceWrite`, `writeResource`). No regression in `wrapper()`'s source-based cascade, the null-safe revision projection, or the removed `Objects.requireNonNull` (confirmed absent).

- 🟠 New regression test's core assertion is very likely incorrect (see inline comment on `SyncFlowsTest.java:769`). `wrapper()`'s ADDED/UPDATED/UNCHANGED decision is derived from `flowBeforeUpdate`, which comes from the independent ZIP-export-based `fetchResources()` call — a code path entirely unaffected by `forceGetFlowStatus`, which only intercepts the single-flow `GET /flows/{ns}/{id}` handler. Since nothing changed in Kestra or Git between the baseline run and the forced-500 dry run, `flowBeforeUpdate` will be non-null with source matching the git source, which makes `ADDED` (which requires `flowBeforeUpdate == null`) unreachable — the actual outcome should be `UNCHANGED`. This matches the PR author's own caveat that the test "has NOT actually been observed passing."
- 🟡 The justification comment on `SyncFlows.java:402-403` ("a permissions/API failure here would otherwise be silently misreported as the flow being ADDED") overstates the actual impact given the above: a non-404 failure inside `simulateResourceWrite`/`writeResource` cannot flip the diff's `syncState` to `ADDED` for a pre-existing flow (that's decided independently from the export-based before-state), it can only cause the *projected `revision`* in the diff/imported-flow object to fall back to `1` (or to the parsed-but-unset value) instead of the real current/next revision. Worth tightening the comment so the next reader understands the actual (narrower) blast radius of this failure mode — this is a documentation-accuracy nit, not a functional defect in the production code path itself.

### Security (OWASP Top 10:2025 + KPS) — 0 findings
✅ No issues found. The new `runContext.logger().warn(...)` call logs namespace, flow id, HTTP status and `e.getMessage()` only — no credentials, tokens, or response bodies. `MockKestraApiServer`'s cosmetic reformatting (brace style, removed unused `InputStream` import, added `ConcurrentHashMap` for `forcedGetFlowStatuses`) removed nothing functional — confirmed by re-reading every hunk of `31bc53f`.

### Performance — 0 findings
✅ No issues found. Precompiling `REVISION_LINE_PATTERN` as a `static final Pattern` is a straightforward, correct fix for the cycle-1 finding (per-call `Matcher` instantiation inside a static method is the standard safe pattern — no shared mutable `Matcher` state). No new allocations, loops, or blocking calls introduced by the `RunContext` threading change.

### CI note
Per the reviewer brief, `check / main`'s prior red run (405 on `POST /api/v1/main/dashboards` in `PushDashboardsContainerTest`/`SyncDashboardsContainerTest`) is attributed to a floating `kestra/kestra:develop-slim` image re-push unrelated to this diff — agreed, nothing in `31bc53f` touches dashboards code paths. At review time `check / main` shows `pending` on the latest push; recommend confirming the same two tests (and not a new failure) once it completes, and separately confirming whether `dryRun_whenFlowApiFailsWithNon404Status_shouldLogWarningAndNotThrow` passes once `GH_PERSONAL_TOKEN` is available in CI, given the finding above.

**Verdict:** REQUEST CHANGES


*(Note: this review is posted with the COMMENT event because GitHub does not allow REQUEST_CHANGES/APPROVE on one's own pull request; treat the Verdict line below as authoritative.)*

*(Also: a stray test review ("test body", COMMENT) was posted earlier while debugging the review-posting tooling for this cycle; please disregard it — it carries no findings. This review is the authoritative one for cycle 2.)*

jymaire:
## QA report — PASS

**Edition**: EE | **Docker tag**: `develop` | **Plugin**: `plugin-git 3.1.2-SNAPSHOT` (this branch)
**Instance**: isolated throwaway container on `:8090` (H2, dedicated volume, no `/tmp` mount)

Driven via the REST API rather than the browser. The `SyncFlows` task authenticates through `kestra.tasks.sdk.authentication.*` (or a task-level `auth:` block) — worth knowing for anyone reproducing this, since without it the task fails with `401` on `/flows/export/by-query` before reaching any sync logic.

### Flow under test

```yaml
id: qa_sync_twice
namespace: qa.git327
tasks:
  - id: sync
    url: https://github.com/brian-mulier-p/test-validation
    branch: master
    gitDirectory: flows
    targetNamespace: qa.git327.imported
    dryRun: false
```

### Results

Revision of the imported flow `qa.git327.imported.hello_world` is used as the behavioural probe — it only advances when the task actually re-imports.

| # | Scenario | Result | Flow revision | Verdict |
|---|----------|--------|---------------|---------|
| 1 | First sync (`dryRun: false`) | SUCCESS | → 1 | ADDED ✅ |
| 2 | **Second sync, nothing changed** | **SUCCESS** | **stays 1** | **UNCHANGED ✅ — this is #327; previously threw the NPE here** |
| 3 | `dryRun: true`, run twice | SUCCESS ×2 | n/a | latent dry-run NPE path clear ✅ |
| 4 | Flow edited in Kestra (1→2), then re-synced | SUCCESS | → 3 | UPDATED ✅ |
| 5 | Fifth run, nothing changed | SUCCESS | stays 3 | UNCHANGED ✅ |

### Why these results are meaningful

The fix replaced revision arithmetic with source-text equality, so QA had to rule out both degenerate outcomes:

- **If it always reported UNCHANGED** (the failure mode of the bare null-guard rejected in the plan), scenario 4 would have left the flow at revision 2 — the drift would never have been corrected. It went to 3.
- **If it always reported UPDATED**, scenarios 2 and 5 would have bumped the revision on every run. They did not.

No NPE in any run. Sync state is correctly discriminated in both directions.

### Not covered

Namespace-file and dashboard sync paths were out of scope. Dashboard functionality is separately broken on the current `develop-slim`/`develop` images (405 on `POST /dashboards`), unrelated to this PR — see the earlier comment.

jymaire (commented):
## Kestra Plugin Code Review (cycle 3, final)

NPE on repeated `SyncFlows` runs (issue #327) is fixed: sync-state derivation moved from nullable-revision comparison to normalized source-text comparison, and the authoritative revision is fetched via the single-flow API instead of the ZIP export (which never carries a `revision`). Regression coverage added for: re-run-twice, dry-run with a manually-edited flow, and non-404 API-failure handling.

### Cycle-3 focus: does `e84e671`'s test rework preserve intent/coverage?

- **Pebble rendering coverage**: not dropped for the class. Only `dryRun_whenFlowApiFailsWithNon404Status_shouldLogWarningAndNotThrow` switched to `Property.ofValue(...)`; the other 5+ tests in `SyncFlowsTest` (`runTwice_shouldReportUnchangedOnSecondRun`, `dryRun_afterManualEditInKestra_...`, `runTwice_withFlowManuallyEditedInKestra_...`, `wrapper_withNullRevisionOnEitherSide_...`, plus the pre-existing cases) still exercise `Property.ofExpression("{{...}}")` end to end. No class-wide regression.
- **`mockRunContext(TENANT_ID, ...)` vs the old `runContextFactory.of(Map)`**: `TestsUtils.mockFlow` derives `flowInfo().namespace()`/`.id()` from the calling test's class/method name rather than the fixed `NAMESPACE`/`FLOW_ID` constants used by the old helper. Traced this through `mustKeep()` (self-flow exclusion) — it only matters when a synced flow's id/namespace matches the running flow's, which never happens for this test's fixture (`unchanged-flow` in `my.namespace` vs. a namespace derived from the test class name). `tenantId` is explicitly passed and stays `"main"` either way. No behavioral difference for what this test asserts.
- **WARN assertion specificity**: verified the predicate (`WARN` level + message contains namespace + flow id + `"500"`) can only match the one warning emitted by `fetchFlowFromApi`'s catch block for this exact flow/status — the baseline sync (before `forceGetFlowStatus` is set) cannot produce a matching WARN, and no other flow in the fixture has a forced failure. `hasSize(1)` is a safe, non-vacuous assertion.
- **`.id("sync-flows")` / `.type(SyncFlows.class.getName())`**: not a workaround. `TestsUtils.mockRunContext` builds a real `TaskRun` from the `Task`, which needs a resolvable id/type — exactly the pattern already used in `SyncTest.reconcile_DryRun_ShouldDoNothing` (`Sync.builder().id("reconcile").type(Sync.class.getName())...`). Every real flow always sets `id`/`type` on tasks at parse time, so this doesn't paper over anything a real user would hit.
- **Vacuity check re-confirmed by reasoning**: `fetchFlowFromApi`'s WARN log is the only thing this test protects; source-based `wrapper()` comparison is correctly unaffected by the single-flow API failure since `flowBeforeUpdate` comes from the namespace ZIP export (`fetchFlowsFromKestra`), not from the failing call — matches the commit message's stated reasoning and the actual call graph in `AbstractSyncTask` (`beforeUpdateResourcesByUri` is populated via `fetchResources` → `fetchFlowsFromKestra`, independent of `fetchFlowFromApi`).
- **Cumulative diff re-check**: `wrapper()`'s source-based comparison, the null-safe dry-run revision projection (`currentFlow.getRevision() != null ? ... : 1`), the `Objects.requireNonNull` removal (only a null-safe `Objects.equals` remains), and the precompiled `REVISION_LINE_PATTERN` are all present and consistent with cycle-1/2 feedback. `Flow.getSource()` returning `null` by design and `FlowWithSource` overriding it means the `Flow`-typed `wrapper()` parameters still dispatch to the real source via virtual dispatch — no NPE/ClassCastException risk there.

### Kestra Guidelines — 1 new finding (non-blocking)
- 🟡 See inline comment on `SyncFlows.java` — `simulateResourceWrite`'s "new flow" fallback (revision `1`) is indistinguishable from "fetch failed for a non-404 reason", which can misreport the projected revision in a dry-run diff.
- 🟢 `MockKestraApiServer.java` mixes the functional `handleGetFlow` addition with unrelated brace/ternary reformatting (lambda body, `handleSearchDashboards`, `handleCreateDashboard`, `extractYamlFromMultipart`). Pre-existing since an earlier commit in this PR (already implicitly accepted in cycle 1/2); flagging only as an optional cleanup, not asking for a respin.

### Security (OWASP Top 10:2025 + KPS) — ✅ No issues found
No secret handling changes, no new HTTP surface beyond the existing SDK client call, WARN log only includes namespace/flow id/status/message — no credential leakage.

`REVISION_LINE_PATTERN` now a precompiled `static final`. `fetchFlowFromApi` no longer buffers/unzips a whole export for a single flow lookup — strictly better than the code it replaces.

### CI
`check / main` failures are `PushDashboardsContainerTest`/`SyncDashboardsContainerTest` 405s against `POST /api/v1/main/dashboards`, consistent with the reported floating `kestra/kestra:develop-slim` image issue — unrelated to this diff, not disputed.

**Verdict:** APPROVE
_(Posted as a COMMENT review — GitHub does not allow self-approving your own PR via the API; the author is also the reviewer's authenticated account here.)_

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -208,25 +209,33 @@ protected Flow simulateResourceWrite(RunContext runContext, String renderedNames
         // Simulate: parse the source and determine if it differs from the current Kestra state
         var parsedFromSource = YamlParser.parse(flowSource, Flow.class);
         var tenantId = runContext.flowInfo().tenantId();
-        var currentFlow = fetchSingleFlow(kestraClient(runContext), tenantId, parsedFromSource.getNamespace(), parsedFromSource.getId());
+        var currentFlow = fetchFlowFromApi(runContext, kestraClient(runContext), tenantId, parsedFromSource.getNamespace(), parsedFromSource.getId());
 
         if (currentFlow == null) {
             // New flow — revision will be 1 after import
-            return parsedFromSource.toBuilder().revision(1).build();
+            return simulatedFlow(parsedFromSource, flowSource, 1);
```
jymaire, line 221: 🟡 [Guidelines] Ambiguous "new flow" fallback on fetch failure
Problem: `fetchFlowFromApi` returns `null` both when the flow genuinely doesn't exist (404) and when the call fails for any other reason (403/5xx, now logged as a WARN). `simulateResourceWrite` can't tell these apart and always projects revision `1` for the "null" case, so a transient 500 on an existing flow with e.g. revision 7 would report `revision: 1` in the dry-run diff instead of the correct `8` (or an "unknown" marker) — misleading whoever reads that diff. The current regression test happens not to catch this because its baseline flow's real revision is coincidentally `1`.
Fix: propagate the distinction out of `fetchFlowFromApi` (e.g. a sentinel/Optional distinguishing "confirmed absent" vs "fetch failed") and skip/flag the revision projection instead of defaulting to 1 when the failure wasn't a 404.

jymaire: Fixed in d2f4f0c: `fetchFlowFromApi` now returns a private `FlowLookup(FlowWithSource flow, boolean resolved)` record. 404 -> `(null, true)` (confirmed absent), other ApiException -> WARN + `(null, false)` (unresolved). `simulateResourceWrite` only projects revision 1 when `resolved` is true; on an unresolved lookup it returns `FlowWithSource.of(parsedFromSource, flowSource)` with no revision, so the dry-run diff reports an absent revision instead of a fabricated 1. Added a regression test (`dryRun_whenFlowApiFailsWithNon404Status_shouldReportNoRevisionInsteadOfFabricatingOne`) asserting `diff.get("revision")` is null for this case. Ran locally against a real GH_PERSONAL_TOKEN: all 13 SyncFlowsTest tests pass.

src/test/java/io/kestra/plugin/git/SyncFlowsTest.java
jymaire: Verified: this was already corrected by c9fe6a9 before this session. Line 774-775 currently asserts `diff.get("syncState")` is `"UNCHANGED"` (not `"ADDED"`), matching the actual before-state coming from the namespace export rather than the failing single-flow GET. Re-ran the full SyncFlowsTest suite locally against the real Gitea/GH fixtures (13 tests, 0 failures) to confirm this assertion holds, including a new sibling test that exercises the revision-projection defect this thread originally flagged. Resolving as stale.

src/main/java/io/kestra/plugin/git/SyncFlows.java
jymaire: Verified: the wording was already tightened by c9fe6a9, and d2f4f0c updated it once more to match the final FlowLookup-based behavior. It now reads: "Only 404 means 'flow does not exist'; any other status is a real API/permissions failure that leaves the lookup unresolved, so callers must not assume the flow is absent." This accurately reflects that the effect is a wrong/absent projected revision, not a misreported ADDED state. Resolving as stale.

jymaire:
### CI update for \`d2f4f0c\`

The final commit reproduces the **same two dashboard failures** and nothing else:

| | \`e84e671\` | \`d2f4f0c\` |
|---|---|---|
| Total | 103 | 105 |
| Passed | 101 | 103 |
| Failed | 2 | 2 |

Failures are identical in both runs — `PushDashboardsContainerTest` and `SyncDashboardsContainerTest`, both on the same `405 Method [POST] not allowed for URI [/api/v1/main/dashboards]`. Neither touches any code path in this diff, which is confined to `SyncFlows.java` and `SyncFlowsTest.java`.

The pass count tracking the total (+2 / +2) shows every added test ran and passed.

One caveat on the analysis in the note above, stated plainly: `main` has **not** been re-run since the `kestra/kestra:develop-slim` re-push, so there is no observed red run on `main` to point at. The attribution rests on the image-push timing plus the fact that the failing tests are in a disjoint code path — not on a reproduction on `main`. Re-running `main` would confirm it, at the cost of showing `main` as red.

All other checks are green: CodeQL, both Analyze jobs, submit-gradle, OpenTelemetry Export Trace.

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -454,6 +470,11 @@ public URI diffFileUri() {
     public static class SyncResult extends AbstractSyncTask.SyncResult {
         private String flowId;
         private String namespace;
+
+        @Schema(
+            title = "Flow revision",
+            description = "Revision of the flow after sync. May be absent if it could not be resolved from the Kestra API."
```
fdelbrayelle: 🟢 [Guidelines] Schema description not a text block
Problem: Per the plugin's Documentation Formatting Rules, @Schema(description = ...) must use a Java text block ("""), not a single-line string literal. Consistent with most of the rest of this file's pre-existing style (also non-compliant), so non-blocking here, but worth aligning on new fields going forward.
Fix: description = """\n    Revision of the flow after sync. May be absent if it could not be resolved from the Kestra API.\n    """

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -376,31 +393,30 @@ private List<Flow> fetchFlowsFromKestra(KestraClient kestraClient, RunContext ru
...
+        } catch (ApiException e) {
+            if (e.getCode() == 404) {
+                return new FlowLookup(null, true);
             }
-            return null;
-        } catch (Exception e) {
-            return null;
+            // Only 404 means "flow does not exist"; any other status is a real API/permissions failure that
+            // leaves the lookup unresolved, so callers must not assume the flow is absent.
+            runContext.logger().warn(
```
fdelbrayelle, line 412: 🟡 [Guidelines] Non-404 API failures degrade silently to a per-flow WARN only
Problem: When the single-flow lookup fails for a non-404 reason, the run still completes and reports the flow as UNCHANGED (source-comparison-based) with only a scattered per-flow WARN log line — there's no aggregate signal in SyncFlows.Output. If the underlying permission/outage issue affects every flow in the namespace (e.g. the calling token lost read access to this specific endpoint), a user watching only the diff output could easily miss it.
Fix: Consider tracking how many lookups were unresolved during a run and surfacing that count (or a boolean flag) on Output, so a systemic failure is visible without having to grep logs.

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -208,25 +209,41 @@ protected Flow simulateResourceWrite(RunContext runContext, String renderedNames
...
             return currentFlow;
         }
 
         // Changed — simulate next revision
-        return parsedFromSource.toBuilder().revision(currentFlow.getRevision() + 1).build();
+        int projectedRevision = currentFlow.getRevision() != null ? currentFlow.getRevision() + 1 : 1;
+        return simulatedFlow(parsedFromSource, flowSource, projectedRevision);
+    }
+
+    private record FlowLookup(FlowWithSource flow, boolean resolved) {
```
fdelbrayelle: 🟢 [Guidelines] Nested record placement
Problem: FlowLookup is declared mid-class, immediately after simulateResourceWrite, interrupting the main RunnableTask lifecycle methods (simulateResourceWrite -> writeResource -> wrapper -> fetchResources).
Fix: Purely stylistic — consider moving FlowLookup next to the other nested types (Output, SyncResult) at the bottom of the file for readability.

fdelbrayelle (approved):
## Kestra Plugin Code Review

Issue #327 reports an NPE when running `SyncFlows` more than once against the same target (both dry-run and non-dry-run, per the issue's flow snippet). The fix removes the nullable-revision dereference from `wrapper()` in favor of normalized-source comparison, and replaces the ZIP-export-based single-flow lookup with the authoritative `flows().flow(...)` REST call. Both the dry-run and non-dry-run repeated-run scenarios are covered by new regression tests (`runTwice_shouldReportUnchangedOnSecondRun`, `SyncFlowsContainerTest#syncFlowsTwice_shouldReportUnchangedOnSecondRun`). Requirement met.

### Kestra Guidelines — 4 findings
- 🟡 Suggestion [DRY] `SyncFlow.java` (sibling task, not touched by this PR) still has the exact bug this PR fixes: its dry-run path at `SyncFlow.java:161-166` catches any `ApiException` from `flows().flow(...)` (401/403/5xx included) and treats it as "flow does not exist," fabricating `projectedRevision = 1`. This is the same false-negative that `SyncFlows.fetchFlowFromApi` now explicitly guards against (404-only). Since the PR description itself references `SyncFlow` as using "the same" endpoint, consider extracting the shared "resolve single flow / distinguish 404-absent from real API failure" logic (and the duplicated `toNamedTempFile` helper) into a common place so `SyncFlow` gets the fix too, or file a follow-up issue so it isn't lost.
- 🟢 Nit — see inline comment on `SyncResult.revision`'s `@Schema`.
- 🟡 Suggestion — see inline comment on the non-404 warning path.
- 🟢 Nit — see inline comment on `FlowLookup` placement.

- Verified `kestraClient.flows().flow(namespace, flowId, tenantId, true, null, false)` matches the SDK's actual `flow(namespace, id, tenant, source, revision, allowDeleted)` signature (checked against `kestra-api-client:1.3.2` sources) — parameters are correctly ordered, no accidental `allowDeleted=true` or wrong-field mix-up.
- No secrets, credentials, or new external inputs touched by this diff. The WARN log at `SyncFlows.java:415-418` only includes namespace/flowId/status/message — no token leakage.
- No injection, deserialization, or access-control surface introduced.

### Performance — 0 issues found
- `REVISION_LINE_PATTERN` is now a precompiled `static final Pattern` instead of being recompiled per call.
- Replacing the per-flow ZIP-export lookup (`fetchSingleFlow`, one archive built server-side per flow) with a direct `GET /flows/{namespace}/{id}` call is a net performance improvement, not a regression, and keeps the same one-call-per-flow cardinality as before.

### Other notes (non-blocking)
- CI on this PR (`check / main`) is currently red, but the 2 failing tests (`PushDashboardsContainerTest`, `SyncDashboardsContainerTest`, both failing on `405 Method not allowed` for `POST /dashboards`) are in files untouched by this diff and unrelated to `SyncFlows`/flows. `main`'s own recent scheduled runs are green, so this looks like environment drift (e.g. container image) rather than something introduced here — worth a quick confirmation before merge that it's pre-existing/unrelated, but it does not reflect a defect in this PR's code.
- Good test discipline overall: the new `wrapper_withNullRevisionOnEitherSide_shouldNotThrowAndCompareBySource` test exercises `wrapper()` directly and needs no network, and the non-404-failure tests correctly use `TestsUtils.mockRunContext` (documented in-line) after discovering via CI that the Map-based `runContext()` helper doesn't wire logs to the log queue — good catch, and useful precedent for future tests in this file.

**Verdict:** APPROVE

src/main/java/io/kestra/plugin/git/SyncFlows.java
jymaire: Done in dac4a39 — `FlowLookup` moved down next to `Output` and `SyncResult`.

src/main/java/io/kestra/plugin/git/SyncFlows.java
jymaire: Done in dac4a39 — switched to a text block.

src/main/java/io/kestra/plugin/git/SyncFlows.java
jymaire: Good catch — agreed that a per-flow WARN is too quiet for a systemic failure. Filed as #336 rather than folding it in here, since it adds a key to the task's public output surface (needs its own `@Schema` docs pass and test) and this PR is already approved and QA'd against #327.

jymaire:
Thanks for the review @fdelbrayelle — all four findings handled.

**Applied in `dac4a39`:**
- 🟢 `SyncResult.revision` `@Schema` description now uses a text block.
- 🟢 `FlowLookup` moved down next to `Output` / `SyncResult` so it no longer interrupts the lifecycle methods.

**Filed as follow-ups**, to keep this PR scoped to #327 (already approved and QA'd):
- 🟡 #335 — `SyncFlow` still swallows 401/403/5xx as "flow does not exist" and fabricates revision 1. Same defect class as #327. The issue covers your shared-helper suggestion, since the duplicated `toNamedTempFile` argues for extraction over copying the fix a second time.
- 🟡 #336 — surface unresolved lookups on `SyncFlows.Output` instead of per-flow WARN logs only. Held back because it adds a key to the public output surface, which wants its own docs pass and test.

On the red CI: confirmed the two failures (`PushDashboardsContainerTest`, `SyncDashboardsContainerTest`) reproduce identically on `d2f4f0c` and nothing else moved — test totals went 103 → 105 with passes 101 → 103, failures constant at the same two. One correction to your note, though: `main`'s last run was `2026-09-04T04:14:07Z`, which is *before* the `kestra/kestra:develop-slim` re-push at `08:11:41Z`, so there is no green *post-image-push* run on `main` to compare against — the attribution rests on the image timing plus the disjoint code path, not on a reproduction. I didn't re-run `main` since that would leave it showing red.

Issue #166, opened 2025-08-06
[PushDashboard] 'dashboard' property doesn't work/filter with custom dashboard id


Malaydewangan09:
### Describe the issue

- Add new dashboard with an id.
- Run the flow below:
- Notice the any pattern/single value in `dashboards` property doesnt filter with custom id, instead works with internal id.

```yaml
id: push_dashboards
namespace: qa-git

tasks:
  - id: push_dashboards
    type: io.kestra.plugin.git.PushDashboards
    version: 1.0.0-SNAPSHOT
    url: https://github.com/Ben8t/qa-git-task
    username: Ben8t
    password: "{{ secret('GITHUB_TOKEN')}}"
    gitDirectory: dashboards
    branch: malay
    dashboards: somedashboard
    commitMessage: "Add dashboards from flow: {{ flow.id }}"
```    

<img width="543" height="250" alt="Image" src="https://github.com/user-attachments/assets/454f7f39-4687-4ef5-8642-cd06bb8268e6"/>

### Environment

- Kestra Version: develop

fdelbrayelle:
Linked to https://github.com/kestra-io/kestra-ee/issues/4484

Malaydewangan09:
Works fine after Brain’s fix on `Dashboards` [here](https://github.com/kestra-io/kestra/pull/10912) 

flow:
```yaml
id: squirrel_749019
namespace: company.team

tasks:
  - id: hello
    type: io.kestra.plugin.git.PushDashboards
    dashboards: my*
    url: https://github.com/Malaydewangan09/kestra-git
    username: Malaydewangan09
    password: "{{ secret('GITHUB_ACCESS_TOKEN') }}"
    branch: dashoard
    commitMessage: "add mydashboard from {{ flow.namespace ~ '.' ~ flow.id }}"
```

<img width="1440" height="795" alt="Image" src="https://github.com/user-attachments/assets/62aff4b2-dfb1-473f-8b87-84df8bb39492" />

PR #340, opened 2026-09-07
fix(dashboards)!: remove dashboard support from the OSS plugin

### LINKED ISSUE #334: SyncDashboards fails on OSS: dashboard write API moved to EE (kestra#18971)
## Summary

`io.kestra.plugin.git.SyncDashboards` always fails on OSS Kestra: kestra core PR [kestra-io/kestra#18971](https://github.com/kestra-io/kestra/pull/18971) (`refactor(dashboards)!: move the custom-dashboard backend to the ee`) removed the dashboard create/update/delete API routes from OSS and moved them to EE. `SyncDashboards.writeResource()`/`deleteResource()` still call those routes, so the task throws on every run against a stock OSS instance. `PushDashboards` (read-only) is unaffected. Companion EE-side issue: [kestra-io/plugin-ee-git#166](https://github.com/kestra-io/plugin-ee-git/issues/166).

## Actual Behaviour

`SyncDashboards` throws `KestraRuntimeException: Failed to save dashboard <id>` (wrapping a `405 Method Not Allowed`) on any write (create or update). `deleteResource()` fails the same way when `delete: true`. This is not intermittent — every OSS instance built from `develop` after 2026-09-04 08:11 UTC (when `#18971` landed on the `kestra/kestra:develop-slim` floating tag) hits it.

Root cause, verified directly against `kestra/kestra:develop-slim`:

```
$ curl -u admin@kestra.io:Admin1234 -X POST http://localhost:18080/api/v1/main/dashboards -d '{}'
HTTP/1.1 405 Method Not Allowed
Allow: HEAD, GET
{"type":"https://kestra.io/docs/api-reference/problems/method-not-allowed","title":"Method not allowed","status":405,"detail":"Method [POST] not allowed for URI [/api/v1/main/dashboards]. Allowed methods: [HEAD, GET]","instance":"/api/v1/main/dashboards"}
```

The OSS `DashboardController` (`kestra` repo) now only serves `GET`/`HEAD` for the bundled default dashboards; `POST`/`PUT`/`DELETE` were moved to the EE-only subclass by #18971.

This is also why `PushDashboardsContainerTest` and `SyncDashboardsContainerTest` started failing in CI on unrelated PRs once the floating `develop-slim` tag picked up #18971 — `SyncDashboardsContainerTest` is testing a task that can no longer work on OSS.

## Expected Behaviour

Either:
- `SyncDashboards` (and `SyncDashboardsContainerTest`) is removed from OSS `plugin-git` and reintroduced in `plugin-ee-git` (tracked in [kestra-io/plugin-ee-git#166](https://github.com/kestra-io/plugin-ee-git/issues/166)) — matching the OSS/EE split #18971 made for the underlying API, or
- if there's a reason to keep a stub here, `SyncDashboards` should fail fast with a clear "dashboard persistence is an Enterprise feature" error instead of an opaque 405-wrapped `KestraRuntimeException`.

`PushDashboards` (read-only, `GET`-based) is unaffected by #18971 and should stay in OSS as-is — no change needed there.

## Reproducer

```yaml
id: git_sync
namespace: company.team

tasks:
  - id: sync_dashboards
    type: io.kestra.plugin.git.SyncDashboards
    gitDirectory: _dashboards
    branch: dashboards
    username: kestra-git-user
    url: https://github.com/my-company/my-repo
    dryRun: false
```

Steps:
1. Run this flow (or `SyncDashboardsContainerTest`) against any OSS `kestra/kestra` image built after #18971 landed on `develop` (e.g. current `kestra/kestra:develop-slim`).
2. Have at least one dashboard YAML file under `_dashboards/` in the target branch.
3. Task fails on the create/update call.

## Logs / Stack Trace

```
io.kestra.core.exceptions.KestraRuntimeException: Failed to save dashboard <id>
Caused by: [HTTP client exception] 405 Method Not Allowed for POST /api/v1/main/dashboards
  {"type":"https://kestra.io/docs/api-reference/problems/method-not-allowed","title":"Method not allowed","status":405,"detail":"Method [POST] not allowed for URI [/api/v1/main/dashboards]. Allowed methods: [HEAD, GET]","instance":"/api/v1/main/dashboards"}
```

(Reconstructed from `SyncDashboards.writeResource()`'s catch block, `plugin-git/src/main/java/io/kestra/plugin/git/SyncDashboards.java:157-161`, plus the raw 405 response reproduced directly above via `curl` against the container.)

## Environment

- **Kestra version**: 2.1.0-SNAPSHOT (`kestra/kestra:develop-slim`, built from `develop` after commit `677d2656d5`)
- **Plugin version**: `plugin-git` 3.1.2-SNAPSHOT (current `develop`)
- **Deployment**: OSS self-hosted (verified via `docker run kestra/kestra:develop-slim server local`)

## Additional Context

- Root cause commit: `677d2656d5` — [kestra-io/kestra#18971](https://github.com/kestra-io/kestra/pull/18971), a deliberate breaking change (`!`) moving dashboard persistence to EE. Verified `kestra` and `kestra-ee` are both correct and in sync on `develop` — this is not a bug in either of those repos.
- Companion issue (EE-side task to restore the capability): [kestra-io/plugin-ee-git#166](https://github.com/kestra-io/plugin-ee-git/issues/166).
- Prior related issue on this exact task: plugin-ee-git#16 (customer-reported, closed 2025-08-28) — confirms real usage of `SyncDashboards`, worth preserving in the correct (EE) repo rather than dropping outright.

---
*[View as Artifact](https://claude.ai/code/artifact/494ca0b5-c8bd-4588-b248-c03e66d8c699)*

Malaydewangan09:
Closes #334.

Dashboards moved to EE in kestra#18971. On OSS 2.0.0 `POST /dashboards` returns 405 and `searchDashboards` is hardcoded to return an empty page, so nothing dashboard-related can work on an OSS instance.

Takes #334's first option (full removal) rather than the fail-fast stub:

- **`SyncDashboards` and `PushDashboards` deleted.** Sync failed with an opaque 405 mid-run; Push committed nothing at all, since OSS has no dashboards to list.
- **`TenantSync`'s dashboard sync removed.** It routed the 405 into `handleInvalid`, whose default is FAIL, so a whole tenant sync died on the first dashboard in Git with a misleading "Invalid syntax for DASHBOARD" error. `_global/dashboards` is now simply ignored.
- **Dead code removed:** `Kind.DASHBOARD`, the mock server's dashboard handlers, `DashboardUtils`, `KestraContainerTestDataUtils.createDashboard`, and the four dashboard test classes.
- **Regression test added:** `shouldIgnoreGlobalDashboardsDirectoryWhenDiscoveringNamespaces` asserts a Git tree containing `_global/dashboards/*.yaml` yields only real namespaces, so dashboard files are ignored instead of failing the sync.

EE is unaffected: plugin-ee-git ships its own `io.kestra.plugin.git.TenantSync` and `NamespaceSync` with full dashboard support, has its own `AbstractGitTask`, and does not depend on this plugin.

**Breaking change.** OSS flows referencing `io.kestra.plugin.git.SyncDashboards` or `PushDashboards` will no longer load. Needs a release note.

**Gate:** plugin-ee-git#166 should land before this releases. These tasks currently still work on EE, so until #166 provides replacements, EE users lose them too, and plugin-ee-git#16 shows real customer usage.


src/main/java/io/kestra/plugin/git/TenantSync.java
```diff
@@ -21,7 +20,6 @@
 import org.eclipse.jgit.transport.PushResult;
 import org.eclipse.jgit.transport.RemoteRefUpdate;
 
-import com.fasterxml.jackson.core.type.TypeReference;
```
fdelbrayelle: 🟢 [Guidelines] Double blank line
Problem: Removing the `TypeReference` import left two consecutive blank lines (22-23) between the jgit imports and the io.kestra imports.
Fix: Collapse to a single blank line.

src/main/java/io/kestra/plugin/git/TenantSync.java
```diff
@@ -296,11 +292,6 @@ public Output run(RunContext runContext) throws Exception {
             );
         }
 
-        planDashboards(
-            runContext, kestraClient, baseDir,
-            rSourceOfTruth, rWhenMissingInSource,
-            rOnInvalidSyntax, rDryRun, diffs, apply
-        );
```
fdelbrayelle: 🟢 [Guidelines] Double blank line
Problem: Removing the `planDashboards(...)` call left two consecutive blank lines (294-295) between the namespace loop and `addPattern`.
Fix: Collapse to a single blank line.

fdelbrayelle (changes requested):
## Kestra Plugin Code Review

### Business Requirements — partial
This PR ("Refs #334") removes dashboard sync from `TenantSync` because `kestraClient.dashboards().createDashboard()/deleteDashboard()` 405s on OSS since kestra#18971 moved dashboard persistence to EE. The fix is correct and well-targeted for `TenantSync` itself.

However, **issue #334 is specifically about `SyncDashboards`** (`SyncDashboards.writeResource()`/`deleteResource()`, per the issue's stack trace and reproducer), not `TenantSync`. `SyncDashboards.java` is untouched by this PR and still calls `kestraClient.dashboards().createDashboard()` (line ~158) and `.deleteDashboard()` (line ~104) directly — it will still throw the same 405-wrapped `KestraRuntimeException` on any OSS instance. The issue's own "Expected Behaviour" section asks for either removing `SyncDashboards` from OSS or making it fail fast with a clear EE-feature message; neither happened here. `PushDashboards` (read-only, correctly out of scope) is unaffected either way.

Recommend either folding a `SyncDashboards` fix into this PR, or explicitly scoping the PR title/description to "partially addresses #334 (TenantSync only)" and opening/linking a dedicated follow-up for `SyncDashboards` so the tracking issue isn't inadvertently closed by this merge while the originally reported symptom is still live.

- 🟠 **Missing regression test for the fix** — `TenantSyncTest.java` has zero dashboard-related coverage before or after this change (verified: no "dashboard" occurrences in the test file, no `TenantSyncContainerTest`). The bug being fixed (a Git repo containing `_global/dashboards/*.yaml` previously made the whole tenant sync fail via `handleInvalid` → FAIL) has no test proving the fix, and nothing guards against this dashboard-sync code path being silently reintroduced later. Per this repo's test bar, a change like this needs at least one test asserting `TenantSync` succeeds against a repo/namespace that includes dashboard-shaped files (or documenting that they're now simply ignored).
- 🟡 **Dead code left behind: `Kind.DASHBOARD`** — `AbstractGitTask.java`'s `Kind` enum still declares `DASHBOARD`, but after this PR nothing in `src/main/java` ever constructs a `DiffLine` with it (confirmed via `git grep Kind.DASHBOARD` on the PR branch — zero matches outside the enum declaration itself). Since `TenantSync` was the only producer, this constant is now unreachable dead code (YAGNI) and should be removed in this PR or a fast follow-up.
- 🟢 **Leftover double blank lines** from the two removed import/call-site blocks (see inline comments) — cosmetic, but worth a quick cleanup pass since it's a one-line fix in the same diff.

### Security (OWASP Top 10:2025 + KPS) — ✅ No issues found
Pure deletion of a dead code path; no new HTTP calls, credentials, or user-controlled input introduced. Secondary observation (non-blocking, since the code is removed): the deleted `fetchDashboardSourceCode` built its request path as `/api/v1/" + tenantId + "/dashboards/" + encodedId` — `dashboardId` was URL-encoded but `tenantId` was not. No action needed since that code is gone, but worth a separate audit pass if similar tenant-in-path construction exists elsewhere in the plugin.

Removal eliminates a paginated dashboard-search loop plus one extra HTTP round trip per dashboard (`fetchDashboardSourceCode`) from every `TenantSync` run — net positive.

**Verdict:** REQUEST CHANGES

src/test/java/io/kestra/plugin/git/MockKestraApiServer.java
```diff
@@ -45,39 +43,24 @@ public class MockKestraApiServer implements AutoCloseable {
...
         this.server.start();
     }
 
-    /**
-     * Creates and starts a mock server backed by the given repositories.
-     * Call {@link #close()} when the test finishes.
-     */
-    public static MockKestraApiServer start(FlowRepositoryInterface flowRepo, MockDashboardStore dashboardRepo) throws IOException {
-        return new MockKestraApiServer(flowRepo, dashboardRepo);
-    }
```
fdelbrayelle: 🟢 [Guidelines] Leftover double blank line
Problem: Two consecutive blank lines were left behind by the dashboard-handler removal (here, and again at line ~64 before `url()`) — a leftover from deleting the javadoc/overload that used to sit between these members. Purely cosmetic.
Fix: Collapse to a single blank line in both spots.

src/test/java/io/kestra/plugin/git/TenantSyncTest.java
```diff
@@ -165,4 +166,22 @@ private static Method fetchReadGitFilesMethod() throws Exception {
...
+
+    @Test
+    void shouldIgnoreGlobalDashboardsDirectoryWhenDiscoveringNamespaces(@TempDir Path tempDir) throws Exception {
+        Files.createDirectories(tempDir.resolve("my.namespace").resolve("flows"));
+        Path dashboardsDir = Files.createDirectories(tempDir.resolve("_global").resolve("dashboards"));
+        Files.writeString(dashboardsDir.resolve("my-dashboard.yaml"), "id: my-dashboard\n", StandardCharsets.UTF_8);
+
+        var task = TenantSync.builder().build();
+        var method = TenantSync.class.getDeclaredMethod("discoverGitNamespaces", Path.class);
+        method.setAccessible(true);
```
fdelbrayelle, line 178: 🟡 [Guidelines] Regression test bypasses RunContext / public API
Problem: This test reflects into the private `discoverGitNamespaces` method instead of exercising `TenantSync.run()` through `RunContextFactory` like the rest of this test class does (see `shouldFailOnInvalidTaskDefinitionFromKestraExport` above). It proves the discovery helper ignores `_global/dashboards`, but not that a full sync with such a directory present actually completes successfully end-to-end (git add/commit/push path, diff output, etc.).
Fix: Consider a companion (or replacement) test that runs the full task via `RunContextFactory` against a repo containing `_global/dashboards/*.yaml`, using the existing `MockKestraApiServer` pattern, asserting `State.Type.SUCCESS` and no dashboard-related calls/errors. Non-blocking — the current unit test is still a valid, narrower regression guard.

fdelbrayelle (approved):
## Kestra Plugin Code Review

Fresh full-PR review at current HEAD (`9fa0b11`, single commit `fix(dashboards)!: remove dashboard support from the OSS plugin`). Note: the branch was rewritten since the prior reviews — it now takes issue #334's "full removal" option instead of the previously-reviewed "deprecate + fail fast" approach, which resolves most previously open findings by construction.

### Prior open items — status
1. Scope gap on #334 (SyncDashboards 405 on OSS) — **fixed**, and more thoroughly than the previous attempt: `SyncDashboards`/`PushDashboards` are deleted outright rather than stubbed.
2. Missing regression test in `TenantSyncTest.java` — **fixed**: `shouldIgnoreGlobalDashboardsDirectoryWhenDiscoveringNamespaces` added and passing.
3. Dead `Kind.DASHBOARD` enum in `AbstractGitTask.java:554` — **fixed**, enum value removed.
4. Stray blank lines in `TenantSync.java` (~22-23, ~294-295) — **fixed**, verified no consecutive blank lines remain in the file.
5. Dead template-method code in `PushDashboards.java`/`SyncDashboards.java` — **moot/resolved**: both files are deleted entirely in this new approach, so there is no unreachable code to speak of.
6. Missing test coverage for the fail-fast behavior — **moot**: the fail-fast design was abandoned in favor of full removal, so there is no such behavior left to test.
7. Double blank lines in `MockKestraApiServer.java` (~51, ~65) — **still open**, see inline comment.

Verified independently: `git grep` across the full PR tree shows zero remaining references to `SyncDashboards`, `PushDashboards`, `DashboardUtils`, `MockDashboardStore`, or `Kind.DASHBOARD`; `AGENTS.md`, `README.md`, and the plugin doc (`io.kestra.plugin.git.md`) are all clean of stale dashboard mentions; `./gradlew compileJava compileTestJava` succeeds; `TenantSyncTest` (3/3, including the new regression test) passes.

### Business Requirements — mostly met, one scope point needs sign-off
- Root cause of #334 is correctly fixed, matching the issue's "Option 1" (full removal from OSS, reintroduced in `plugin-ee-git`).
- **Scope beyond the issue**: #334's "Expected Behaviour" explicitly states *"`PushDashboards` (read-only, `GET`-based) is unaffected by #18971 and should stay in OSS as-is — no change needed there."* This PR removes `PushDashboards` too. I verified the PR's justification against `kestra-io/kestra#18971`: OSS `DashboardController.searchDashboards` is indeed hardcoded to `PagedResults.of(new ArrayListTotal<>(List.of(), 0))`, so `PushDashboards` would silently no-op forever on OSS (never enumerates anything to push) rather than fail loudly — a legitimate reason to remove it too. The reasoning is sound and transparently disclosed in the PR description, but it contradicts the linked issue's explicit acceptance criterion, so it's worth an explicit go-ahead from the issue author/product owner before merge. Not a code defect — flagged for awareness, not blocking.
- The PR body correctly calls out the breaking change, gates the release on `plugin-ee-git#166`, and references prior customer usage (`plugin-ee-git#16`) — good practice.

- 🟢 Nit: leftover double blank lines in `MockKestraApiServer.java` (see inline).
- 🟡 Suggestion: new regression test bypasses `RunContext`/public `run()` in favor of reflecting into a private method (see inline).

### Security (OWASP Top 10:2025 + KPS) — no issues found
This diff is almost entirely deletion. The one previously-present risk in this area — `TenantSync.fetchDashboardSourceCode()` building a raw API path by string-concatenating `tenantId` (unencoded) with only `dashboardId` URL-encoded — is now removed entirely along with the rest of the dashboard sync machinery, a net risk reduction. No new HTTP calls, secret handling, or file I/O introduced.

### Performance — no issues found
Pure removal of unreachable/duplicated paged-fetch loops and dashboard diffing logic; no new loops, allocations, or blocking calls introduced.

**Verdict:** APPROVE (non-blocking nits inline; recommend confirming the `PushDashboards` scope expansion with the issue author before merging, per the Business Requirements note above)

PR #342, opened 2026-09-09
refactor: consume shared Git kernel from plugin-git-lib

jymaire:
## Summary

- Deletes `AbstractGitTask`/`AbstractKestraTask`/`AbstractCloningTask`/`AbstractSyncTask`/`AbstractPushTask`,
  `KestraApiAuth`/`KestraApiConnection` and `services.GitService`/`SshTransportConfigCallback`,
  now published from [`plugin-git-lib`](https://github.com/kestra-io/plugin-git-lib) as
  `io.kestra.plugin.git.shared.*`. Depends on it via `api 'io.kestra.plugin:plugin-git-lib:1.0.0-SNAPSHOT'`
  (resolved from `mavenLocal()` until the lib is released).
- Concrete, registered tasks (`Clone`, `Push*`, `Sync*`, `NamespaceSync`, `TenantSync`) stay in this repo
  and now import the shared classes instead of the local ones.
- `Clone` delegates its clone/init-fetch-checkout logic to the lib's new `CloneService` instead of
  duplicating it.
- Drops the now-transitive `jgit`/`kestra-api-client` dependency declarations.
- Test scaffolding (`AbstractGitTest`, `AbstractKestraContainerTest`, `KestraContainerTestDataUtils`,
  `MockKestraApiServer`) is now consumed from the lib's `testFixtures`;
  `KestraApiConnectionTest`/`AbstractKestraTaskAuthTest` are removed here since equivalent coverage now
  lives in the lib.
- Both OSS git tasks gain two new opt-in SSH properties inherited from the merged callback:
  `strictHostKeyChecking` and `knownHosts` (`advanced` group). **Per-edition default, no behavior
  change**: OSS keeps resolving `strictHostKeyChecking` to `false` (unchanged) via the shared lib's
  `defaultStrictHostKeyChecking()` hook; the Enterprise Edition overrides that hook to keep its own
  pre-existing `true` default. Covered by the new `SshHostKeyCheckingDefaultTest`.

part-of: https://github.com/kestra-io/kestra-ee/issues/7117

## Test plan

- [x] `rtk test ./gradlew test` — same 66 pre-existing network/credential-gated failures as the
      unmodified `main` baseline (no GitHub PAT or Gitea container in this environment); test count
      delta (98 → 88) is the 12 tests relocated to `plugin-git-lib` (now passing there) minus the
      2 new `SshHostKeyCheckingDefaultTest` cases added here (both pass)
- [x] `./gradlew compileJava compileTestJava` clean
[...]

PR #344, opened 2026-09-10
fix(SyncFlow): fail when the requested branch does not exist

### LINKED ISSUE #343: Sync tasks silently fall back to the default branch when `branch` does not exist
`SyncFlows` / `SyncNamespaceFiles` with a `branch` that does not exist on the remote do not fail. `GitService.cloneBranch` logs `Branch X does not exist, creating it`, clones the repository's default HEAD and creates the branch locally, so the task syncs the **default branch's** content — and with `delete: true` deletes everything not present there. The execution succeeds.

Creating the missing branch is right for the `Push*` tasks that share this helper. For a sync it turns a typo in `branch`, or a branch someone deleted, into a silent content swap plus deletions, on a green execution.

Measured on Kestra 1.3.37, plugin-git 2.0.11:

```yaml
- id: sync
  url: https://github.com/<org>/<repo>
  branch: no-such-branch
  gitDirectory: flows/maintenance
  targetNamespace: probe.victim
  includeChildNamespaces: true
  delete: true
```

`probe.victim` held `victim-a` and `victim-b`; the default branch (`main`) holds `flows/maintenance/purge-executions.yaml`:

```
INFO  Branch no-such-branch does not exist, creating it
DEBUG {gitPath:null,syncState:"DELETED",flowId:"victim-a",namespace:"probe.victim",revision:1}
DEBUG {gitPath:null,syncState:"DELETED",flowId:"victim-b",namespace:"probe.victim",revision:1}
DEBUG {gitPath:"flows/maintenance/purge-executions.yaml",syncState:"ADDED",flowId:"purge-executions",namespace:"probe.victim",revision:1}
```

Execution state: SUCCESS.

For comparison, a `gitDirectory` that does not exist on the branch fails loudly (`failOnMissingDirectory`, default true) — that is the behaviour I would expect for a missing branch too.

Suggestion: have the sync tasks refuse a branch that is not on the remote — `AbstractSyncTask` calling the existing `GitService.branchExists()` and throwing, or a `failOnMissingBranch` property defaulting to true for sync tasks only.

jymaire:
## Summary

Protects `SyncFlow` from silently syncing the **wrong branch**, the `SyncFlow` half of #343.

When `branch` does not exist on the remote, `GitService.cloneBranch` falls back to the repository's default branch (it creates the missing branch locally from the default HEAD). `SyncFlow` then reads `flowPath` from that default-branch tree, so it can silently import the **default branch's version** of the flow into `targetNamespace` while the execution reports SUCCESS. (Unlike `SyncFlows`/`SyncNamespaceFiles`, `SyncFlow` has no bulk delete, so this is silent wrong-content, not data loss.)

This PR:
- adds a `failOnMissingBranch` property (default `true`) to `SyncFlow`, and
- guards the direct `cloneBranch` call with the shared `GitService.ensureBranchExistsOrFail(...)` primitive, so a missing branch now fails loudly instead of falling back.

The bulk-sync tasks (`SyncFlows`, `SyncNamespaceFiles`) are fixed in the shared kernel via `AbstractSyncTask` and are **not** part of this PR.

## Shared-kernel dependency

`GitService.ensureBranchExistsOrFail(...)` ships in **plugin-git-lib 1.0.2** (kestra-io/plugin-git-lib#6). This branch is rebased onto `main` (which already consumes the shared kernel via #342) and bumps the `plugin-git-lib` `api` + `testFixtures` coordinates from `1.0.0` to `1.0.2`. No longer blocked.

## Test plan

- New regression test `SyncFlowTest#missingBranch_shouldFailInsteadOfFallingBackToDefaultBranch` — syncs a non-existent branch and asserts the task throws `IllegalArgumentException` **and** the mock flow-import endpoint receives zero calls (concrete proof no flow was imported). Uses the class's existing real-remote + mock-Kestra-API harness.
- `./gradlew compileJava compileTestJava` passes against `plugin-git-lib:1.0.2`; `./gradlew test` runs in CI.

Related: kestra-io/plugin-git-lib#6 · fixes #343

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,7 +56,7 @@ dependencies {
     compileOnly group: "io.kestra", name: "script", version: kestraVersion
 
     // shared Git kernel (also brings jgit and kestra-api-client transitively as `api`)
-    api 'io.kestra.plugin:plugin-git-lib:1.0.0'
+    api 'io.kestra.plugin:plugin-git-lib:1.0.2'
 }
 
 
@@ -103,7 +103,7 @@ dependencies {
     testImplementation "org.testcontainers:junit-jupiter:1.21.4"
 
     // shared Gitea/Kestra-container test scaffolding and mock Kestra API server
-    testImplementation testFixtures('io.kestra.plugin:plugin-git-lib:1.0.0')
+    testImplementation testFixtures('io.kestra.plugin:plugin-git-lib:1.0.2')
 }
 
 /**********************************************************************************************************************\
--- a/src/main/java/io/kestra/plugin/git/SyncFlow.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlow.java
@@ -126,7 +126,6 @@ public class SyncFlow extends AbstractKestraTask implements RunnableTask<SyncFlo
         GitService gitService = new GitService(this);
 
         String rBranch = runContext.render(this.getBranch()).as(String.class).orElse(null);
-        // TODO(#343): requires plugin-git-lib >= <next release> for GitService.ensureBranchExistsOrFail
         gitService.ensureBranchExistsOrFail(runContext, rBranch, runContext.render(this.failOnMissingBranch).as(Boolean.class).orElse(true));
         Git git = gitService.cloneBranch(runContext, rBranch, Property.ofValue(Boolean.FALSE));
         Path cloneDir = git.getRepository().getWorkTree().toPath();
--- a/src/main/java/io/kestra/plugin/git/SyncFlow.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlow.java
@@ -104,6 +104,17 @@ public class SyncFlow extends AbstractKestraTask implements RunnableTask<SyncFlo
     @PluginProperty(group = "reliability")
     private Property<Boolean> dryRun = Property.ofValue(Boolean.FALSE);
 
+    @Schema(
+        title = "Fail when the branch does not exist",
+        description = """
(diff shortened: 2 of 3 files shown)


Malaydewangan09:
I unblocked this locally to check it: merged main, published plugin-git-lib#6 as `1.0.1-SNAPSHOT`, bumped the dep. It compiles and the guard works. Also rewrote your test against a local `file://` remote: throws `IllegalArgumentException` with 0 calls to `/flows/import`, and with `failOnMissingBranch: false` it imports the default-branch flow. So the fix is good.

- The branch doesn't contain #342 and `build.gradle` has no `plugin-git-lib` dep at all. So it isn't "doesn't compile against 1.0.0", it's just stale. Merge main first, then bump.
- `// TODO(#343): requires plugin-git-lib >= <next release>` still has the literal placeholder.
- The new test extends `AbstractGitTest`, so it needs `kestra.git.pat` and won't run in a fork PR. That's why CI shows 0 tests ran. The guard fires before the clone, so a temp local repo is enough.
- No test for `failOnMissingBranch: false`.
- The `renderedBranch` to `rBranch` rename and the `.auth(...)` reformat across four unrelated tests are most of the diff. I couldn't find that convention in AGENTS.md, and it makes the fix harder to backport.

Unrelated but you're in the file: `SyncFlow` never calls `namespaceAccessGuard` for `targetNamespace` while `AbstractSyncTask` does, and `git.close()` isn't in a try-with-resources so it leaks on the `FileNotFoundException` path.

Malaydewangan09 (approved):

PR #346, opened 2026-09-11
fix(push): consume plugin-git-lib 1.0.1 with DELETE_ONLY push mode

### LINKED ISSUE #345: PushFlows: delete=true always re-pushes all matched flows, no delete-only mode
## Summary

`PushFlows` (and any `AbstractPushTask` subclass) cannot push a delete-only change. When `delete: true`, deletion of removed flows is staged correctly, but the task *also* unconditionally re-writes and re-stages every flow currently matching `flows` glob, resulting in unrelated adds/changes being pushed alongside the delete.

## Root cause

`AbstractPushTask.run()` (io.kestra.plugin.git.shared, `plugin-git-lib`):
- `contentByPath` (existing/matching flows) is always computed and staged (`AbstractPushTask.java:451-503`).
- `delete` only adds deletion staging via `deleteOutdatedResources` on top (`AbstractPushTask.java:126-196`, `:479-480`).
- There is no option to stage deletions only, without re-syncing all currently matching content.

## Repro

1. Push flows `a` and `b` to Git with `PushFlows`, `flows: **`.
2. Modify flow `a` in Kestra (don't intend to push this change yet).
3. Delete flow `b` in Kestra.
4. Run `PushFlows` with `flows: **`, `delete: true`, intending to only push the deletion of `b`.
5. Observed: commit also includes flow `a`'s unrelated change.

## Also reported (unconfirmed, needs repro)

Same customer report: deleting a single flow (`flows: [<flow_id>]`, `delete: true`) after removing it from Kestra did not stage/push the deletion at all. Code inspection suggests this *should* work (glob is matched against the file stem in `deleteOutdatedResources`), but couldn't reproduce/confirm without the customer's exact plugin/Kestra version. Worth checking as part of this issue — may be a version mismatch or a separate bug.

## Source

Reported via Pylon support ticket #2332: https://app.usepylon.com/support/issues/views/inbox?view=fs&issueNumber=2332

jymaire:
## Summary

Completes the OSS side of #345. The fix itself lives in the shared kernel — `AbstractPushTask` in [plugin-git-lib](https://github.com/kestra-io/plugin-git-lib) ([#8](https://github.com/kestra-io/plugin-git-lib/pull/8), first released as **1.0.1**). This PR:

- bumps the `plugin-git-lib` pin `1.0.0` → **`1.0.2`** in `build.gradle` (both the `api` and `testFixtures` coordinates), pulling in the new `pushMode` property (`SYNC` default, `DELETE_ONLY`);
- adds two end-to-end `PushFlows` tests at the OSS surface, one per need in the originating support report (Pylon #2332).

## Why 1.0.2 rather than 1.0.1

`DELETE_ONLY` shipped in lib 1.0.1, but 1.0.2 is already released and is a **strict superset** — it still carries the `DELETE_ONLY` push mode this PR is built around, and additionally brings two more shared-kernel changes into the OSS plugin:

- **[#6](https://github.com/kestra-io/plugin-git-lib/pull/6)** — sync tasks now fail when the requested branch does not exist on the remote;
- **[#7](https://github.com/kestra-io/plugin-git-lib/pull/7)** — call an unauthenticated Kestra API when `auth.auto` is disabled.

Pinning to the latest released lib keeps OSS from trailing behind and needing an immediate follow-up bump. The two new tests here still exercise only the `DELETE_ONLY` surface.

## Why

Before the fix, `PushFlows` with `delete: true` always re-wrote and re-staged every flow matching `flows`, so a delete-only intent dragged along unrelated modified/new flows. `pushMode: DELETE_ONLY` stages only the removals of flows no longer present in Kestra, writing/staging nothing else.

## Coverage vs the customer's two needs (Pylon #2332)

1. **"Push a delete of a single flow, ignoring all others"** → `deleteOnly_singleFlowGlob_deletesOnlyThatFlow`: a `DELETE_ONLY` push scoped to one flow id removes only that file; another flow — even one changed in Kestra — is untouched.
2. **"Push all deletes without also including all changes and all new flows"** → `deleteOnly_pushesOnlyDeletion_leavesUnrelatedModificationOnBranch` (the exact issue repro): after modifying flow `a` and deleting flow `b` in Kestra, a `DELETE_ONLY` push commits only the deletion of `b`; `a`'s pending change does not ride along.

## Test plan

- [x] `./gradlew compileTestJava` — green against the released lib (resolved from Maven Central).
- [ ] The two new tests are Gitea/PAT-backed (like the rest of `PushFlowsTest`) and run in CI, not in the local sandbox.

## Follow-up

- `plugin-ee-git` needs the same mechanical bump to **1.0.2** so the Enterprise push tasks pick up the mode (and #6/#7).

Closes #345

### DIFF
--- a/src/test/java/io/kestra/plugin/git/PushFlowsTest.java
+++ b/src/test/java/io/kestra/plugin/git/PushFlowsTest.java
@@ -37,6 +37,7 @@ import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.tenant.TenantService;
 import io.kestra.core.utils.IdUtils;
 import io.kestra.core.utils.Rethrow;
+import io.kestra.plugin.git.shared.AbstractPushTask;
 import io.kestra.plugin.git.shared.services.GitService;
 import io.kestra.plugin.git.shared.testkit.AbstractGitTest;
 import io.kestra.plugin.git.shared.testkit.MockKestraApiServer;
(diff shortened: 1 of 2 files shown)


src/test/java/io/kestra/plugin/git/PushFlowsTest.java
```diff
@@ -817,6 +818,138 @@ void defaultCase_ExcludesDraftFlows() throws Exception {
...
+     * With {@code pushMode: DELETE_ONLY} a {@code delete: true} push stages ONLY the removal of a flow deleted from
+     * Kestra; an unrelated change made to another flow must NOT ride along on the commit.
+     */
+    @Test
+    void deleteOnly_pushesOnlyDeletion_leavesUnrelatedModificationOnBranch() throws Exception {
+        runDeleteOnlyScenario(null, "Delete only", false);
+    }
+
+    /**
+     * Pylon #2332 / issue #345, need #1: "push a delete of a single flow, ignoring all other". A DELETE_ONLY push
```
Malaydewangan09: not sure w need these references to Pylon and stuff?

Malaydewangan09 (approved):

src/test/java/io/kestra/plugin/git/PushFlowsTest.java
jymaire: nope, you right

PR #347, opened 2026-09-11
fix(git): reword misleading 404 message in NamespaceSync's namespace check

jymaire:
## Summary

- Reworded `NamespaceSync`'s namespace-existence pre-check: OSS Kestra's `GET /namespaces/{id}` never returns 404 for any id (verified against a live 2.0.0 OSS instance), so the old "the namespace does not exist" message was unreachable in a healthy deployment and mislabeled the real cause whenever it did fire. The new message names both plausible causes (namespace missing, or a `kestraUrl` that doesn't resolve to the Kestra webserver API) and includes the task's resolved `kestraUrl` (or the fallback chain description when unset).
- Tightened three other `ApiException` catch sites that swallowed the HTTP status/body behind a generic message, in a minimal, inline way (no shared helper introduced): `SyncFlows`/`PushFlows` flow export now include `HTTP <status>: <message>` in the wrapped exception, and `SyncFlow`'s dry-run revision lookup now logs a warning naming the status/message for any non-404 failure instead of silently assuming the flow doesn't exist.
- Added a regression test locking down the new 404 message (`NamespaceSyncTest`), using a dedicated local `HttpServer` since the shared `MockKestraApiServer` never 404s on namespace lookups (mirroring real OSS behavior).
- Strengthened `SyncFlowsTest.defaultCase_WithDelete` with an explicit guard: a flow in a sibling namespace (`another.namespace`, not a descendant of the target) must never be returned by the sync or considered for deletion, even with `delete: true` and `includeChildNamespaces: true` — this is the regression guard for the namespace-filter path also confirmed working end-to-end during the manual OSS reproduction (`EQUALS=system` → empty export, `EQUALS=tutorial` → full export).

**Scope note:** this is the OSS-only slice of the fix for the linked issue. The shared-kernel diagnostics helper (wrapping `ApiException` with URL/method/status/body, to be reused consistently across all catch sites without duplicating the formatting) is intentionally deferred to a follow-up `plugin-git-lib` PR + release, per the approved plan. This PR does not bump the `plugin-git-lib` version.

part-of: https://github.com/kestra-io/kestra/issues/19181

## Test plan

- [x] `rtk test ./gradlew test` passes (90/90, including the new `NamespaceSyncTest` and strengthened `SyncFlowsTest` cases) — run locally with a local Gitea fixture (`.github/setup-unit.sh`) and a personal GitHub token exported as `GH_PERSONAL_TOKEN`
- [ ] Manually trigger `NamespaceSync` against a Kestra instance with a deliberately wrong `kestraUrl` and confirm the new message names both causes and shows the resolved URL

---
*[View as Artifact](https://claude.ai/code/artifact/9cca80fe-ff60-44ec-8ecf-02238b873961)*

### DIFF
--- a/src/main/java/io/kestra/plugin/git/PushFlows.java
+++ b/src/main/java/io/kestra/plugin/git/PushFlows.java
@@ -258,7 +258,9 @@ public class PushFlows extends AbstractPushTask<PushFlows.Output> {
         } catch (IOException e) {
             throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
         } catch (ApiException e) {
-            throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
+            throw new KestraRuntimeException(
+                "Failed to export flows from Kestra for namespace " + namespace + " (HTTP " + e.getCode() + "): " + e.getMessage(), e
+            );
         }
     }
 
--- a/src/main/java/io/kestra/plugin/git/SyncFlow.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlow.java
@@ -175,7 +175,14 @@ public class SyncFlow extends AbstractKestraTask implements RunnableTask<SyncFlo
                 FlowWithSource existing = kestraClient.flows().flow(rNamespace, flowId, tenantId, false, null, false);
                 projectedRevision = existing.getRevision() != null ? existing.getRevision() + 1 : 1;
             } catch (ApiException e) {
-                // Flow does not exist yet
+                if (e.getCode() != 404) {
+                    // Only 404 means "flow does not exist yet"; any other status is a real API/permissions
+                    // failure, surfaced here so a misconfigured kestraUrl isn't silently reported as revision 1.
+                    runContext.logger().warn(
+                        "Failed to fetch existing flow {}.{} from the Kestra API (status {}): {} — assuming it does not exist yet for revision projection",
+                        rNamespace, flowId, e.getCode(), e.getMessage()
+                    );
+                }
                 projectedRevision = 1;
             }
 
--- a/src/main/java/io/kestra/plugin/git/SyncFlows.java
+++ b/src/main/java/io/kestra/plugin/git/SyncFlows.java
@@ -387,7 +387,9 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
         } catch (IOException e) {
             throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
         } catch (ApiException e) {
-            throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
+            throw new KestraRuntimeException(
+                "Failed to export flows from Kestra for namespace " + namespace + " (HTTP " + e.getCode() + "): " + e.getMessage(), e
+            );
         }
     }
 
--- a/src/test/java/io/kestra/plugin/git/NamespaceSyncTest.java
+++ b/src/test/java/io/kestra/plugin/git/NamespaceSyncTest.java
@@ -3,6 +3,7 @@ package io.kestra.plugin.git;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.InetSocketAddress;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -15,6 +16,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.sun.net.httpserver.HttpServer;
 
 import io.kestra.core.junit.annotations.KestraTest;
 import io.kestra.core.models.flows.Flow;
@@ -37,6 +39,7 @@ import jakarta.inject.Inject;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.containsString;
 import static org.junit.jupiter.api.Assertions.*;
 
 @KestraTest
@@ -239,6 +242,45 @@ public class NamespaceSyncTest extends AbstractGitTest {
 
         assertTrue(Files.exists(base.resolve("flows/" + flowId + ".yaml")));
         assertFalse(Files.exists(base.resolve("flows/" + draftFlowId + ".yaml")));
+    }
+
+    @Test
+    void namespaceCheck_404_namesBothCausesAndResolvedKestraUrl() throws Exception {
+        // The mock server used by every other test never 404s on GET /namespaces/{id} (as real OSS Kestra
+        // doesn't either), so a dedicated server is needed here to force the 404 branch.
+        var server = HttpServer.create(new InetSocketAddress(0), 0);
+        server.createContext("/api/v1/" + TENANT_ID + "/namespaces/" + NAMESPACE, exchange ->
+        {
+            exchange.getRequestBody().readAllBytes();
+            exchange.sendResponseHeaders(404, -1);
+            exchange.close();
+        });
+        server.start();
+
+        try {
+            var kestraUrl = "http://localhost:" + server.getAddress().getPort();
+            NamespaceSync task = NamespaceSync.builder()
+                .url(Property.ofExpression("{{url}}"))
+                .username(Property.ofExpression("{{pat}}"))
+                .password(Property.ofExpression("{{pat}}"))
+                .branch(Property.ofExpression("{{branch}}"))
+                .namespace(Property.ofExpression("{{namespace}}"))
+                .kestraUrl(Property.ofValue(kestraUrl))
+                .build();
+
+            IllegalArgumentException exception = assertThrows(
+                IllegalArgumentException.class,
+                () -> task.run(runContext())
+            );
+
+            assertThat(exception.getMessage(), containsString(NAMESPACE));
+            assertThat(exception.getMessage(), containsString(TENANT_ID));
+            assertThat(exception.getMessage(), containsString(kestraUrl));
+            assertThat(exception.getMessage(), containsString("does not exist yet"));
+            assertThat(exception.getMessage(), containsString("kestraUrl"));
+        } finally {
+            server.stop(0);
+        }
     }
 
     private RunContext runContext() {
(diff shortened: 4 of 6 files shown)


src/main/java/io/kestra/plugin/git/PushFlows.java
```diff
@@ -258,7 +258,9 @@ private List<FlowWithSource> fetchFlowsFromKestra(KestraClient kestraClient, Run
         } catch (IOException e) {
             throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
         } catch (ApiException e) {
-            throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
+            throw new KestraRuntimeException(
+                "Failed to export flows from Kestra for namespace " + namespace + " (HTTP " + e.getCode() + "): " + e.getMessage(), e
+            );
```
jymaire, line 263: 🟡 [Security/KPS-08] Untruncated upstream response body in thrown exception
Problem: `e.getMessage()` on this SDK's `ApiException` is populated from `EntityUtils.toString(entity)` — the entire non-2xx response body, with no size cap (verified by decompiling kestra-api-client's `ApiClient.processResponse`). Embedding it verbatim here means a verbose or malformed upstream error (e.g. an HTML error page from a proxy sitting in front of a misconfigured kestraUrl) lands unbounded in the task's failure message, visible in execution logs/outputs.
Fix: Truncate before embedding, e.g. `StringUtils.abbreviate(e.getMessage(), 500)` (or a local helper), independent of the deferred shared-kernel diagnostics helper.

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -387,7 +387,9 @@ private List<Flow> fetchFlowsFromKestra(KestraClient kestraClient, RunContext ru
         } catch (IOException e) {
             throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
         } catch (ApiException e) {
-            throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
+            throw new KestraRuntimeException(
+                "Failed to export flows from Kestra for namespace " + namespace + " (HTTP " + e.getCode() + "): " + e.getMessage(), e
+            );
```
jymaire, line 392: 🟡 [Security/KPS-08] Same untruncated response body as PushFlows.java
Problem: identical pattern to `PushFlows.fetchFlowsFromKestra` — `e.getMessage()` carries the full unbounded upstream response body into the thrown `KestraRuntimeException`.
Fix: apply the same truncation fix as the `PushFlows.java` sibling call site so both stay consistent.

src/main/java/io/kestra/plugin/git/SyncFlow.java
```diff
@@ -175,7 +175,14 @@ public Output run(RunContext runContext) throws Exception {
...
                 projectedRevision = existing.getRevision() != null ? existing.getRevision() + 1 : 1;
             } catch (ApiException e) {
-                // Flow does not exist yet
+                if (e.getCode() != 404) {
+                    // Only 404 means "flow does not exist yet"; any other status is a real API/permissions
+                    // failure, surfaced here so a misconfigured kestraUrl isn't silently reported as revision 1.
+                    runContext.logger().warn(
+                        "Failed to fetch existing flow {}.{} from the Kestra API (status {}): {} — assuming it does not exist yet for revision projection",
+                        rNamespace, flowId, e.getCode(), e.getMessage()
+                    );
```
jymaire, line 184: 🟡 [Security/KPS-08] Untruncated response body logged at WARN
Problem: same unbounded-body characteristic as the PushFlows/SyncFlows findings, here logged rather than thrown — lower severity since it doesn't fail the task, but still risks an oversized/opaque log line per invocation.
Fix: consider truncating `e.getMessage()` before logging, for consistency with the other two call sites once fixed.

src/test/java/io/kestra/plugin/git/NamespaceSyncTest.java
```diff
@@ -241,6 +244,45 @@ void kestraToGit_apply_excludesDraftFlows() throws Exception {
         assertFalse(Files.exists(base.resolve("flows/" + draftFlowId + ".yaml")));
     }
 
+    @Test
+    void namespaceCheck_404_namesBothCausesAndResolvedKestraUrl() throws Exception {
+        // The mock server used by every other test never 404s on GET /namespaces/{id} (as real OSS Kestra
+        // doesn't either), so a dedicated server is needed here to force the 404 branch.
+        var server = HttpServer.create(new InetSocketAddress(0), 0);
+        server.createContext("/api/v1/" + TENANT_ID + "/namespaces/" + NAMESPACE, exchange ->
```
jymaire, line 252: 🟢 [Nit] Local variable shadows the class field of the same name
Problem: this method declares `var server = HttpServer.create(...)`, shadowing the test class's `private MockKestraApiServer server` field for the rest of the method body. It compiles and runs correctly (the `finally` block's `server.stop(0)` correctly resolves to the local `com.sun.net.httpserver.HttpServer`), but a future reader skimming this method could momentarily confuse it with the shared mock server field.
Fix: rename the local to something like `fourOhFourServer` or `rawHttpServer` to avoid the shadowing.

jymaire (commented):
## Kestra Plugin Code Review

The diff matches the approved OSS-only scope for kestra#19181: reworded the misleading NamespaceSync 404 message, added HTTP status/body context to three ApiException catch sites inline (no shared helper, deferred as agreed), and added two regression tests. No scope creep.

### Kestra Guidelines — 0 blocking findings, 1 nit
- `NamespaceSync` 404 branch correctly rethrows non-404 `ApiException`s unchanged, and renders `kestraUrl` via `.orElse(null)` consistent with its `@Nullable Property<String>` declaration in the shared kernel — no NPE risk. The described fallback chain (`kestra.tasks.sdk.authentication.url` → `kestra.url` → `http://localhost:8080`) matches `KestraApiConnection.resolve()`'s actual behavior.
- 🟢 Nit — see inline on `NamespaceSyncTest.java`.

### Security (OWASP Top 10:2025 + KPS) — 2 findings, non-blocking
- No basic-auth credential leakage: `ApiException.getMessage()`/`responseBody` are populated from the raw HTTP *response* body only (verified by decompiling `kestra-api-client`'s `ApiClient.processResponse`), never from request headers or credentials.
- 🟡 KPS-08 — the two `(HTTP <status>): <message>` wrappers embed the full, untruncated upstream response body (confirmed via bytecode: the SDK does `EntityUtils.toString(entity)` with no size cap for both `message` and `responseBody`). See inline comments on `PushFlows.java` and `SyncFlows.java`.
- 🟡 Same underlying pattern in `SyncFlow.java`'s new warn log (lower severity — log-only, not user-facing failure text).

### Performance — 0 findings
No new unbounded allocation, blocking loops, or thread-safety issues introduced by this diff. The pre-existing `byte[] zippedFlows` full-buffering in `fetchFlowsFromKestra` predates this PR and is out of scope.

### Tests — verified, not passing for the wrong reason
- `NamespaceSyncTest`: dedicated ephemeral `HttpServer` (random port, torn down in `finally`) is necessary since the shared `MockKestraApiServer` never 404s on namespace lookups — correctly isolated, deterministic. Assertions on namespace/tenant/kestraUrl substrings match the actual rendered values used to build the message.
- `SyncFlowsTest`: the sibling-namespace guard asserts against the *real* `flowRepositoryInterface` state (not the mock's response), so it genuinely detects a namespace-filter regression rather than just echoing the mock. `MockKestraApiServer`'s namespace filter implements real `EQUALS`/`STARTS_WITH` matching, consistent with what SyncFlows sends.

**Verdict:** APPROVE (posted as COMMENT — GitHub does not allow self-approval of your own PR)

src/main/java/io/kestra/plugin/git/PushFlows.java
```diff
@@ -258,7 +258,9 @@ private List<FlowWithSource> fetchFlowsFromKestra(KestraClient kestraClient, Run
         } catch (IOException e) {
             throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
         } catch (ApiException e) {
-            throw new KestraRuntimeException("Failed to export flows from Kestra for namespace " + namespace, e);
+            throw new KestraRuntimeException(
+                "Failed to export flows from Kestra for namespace " + namespace + " (HTTP " + e.getCode() + "): " + ApiErrors.truncate(e.getMessage()), e
```
Malaydewangan09: If i am understnading this, we are adding a message in addition to the stack trace?

Malaydewangan09:
Diagnostics parts are good. Two asks before merge.

The kestraUrl hypothesis is contradicted by the issue: the reporter's 404 body is Kestra's own problem+json with `"instance": "/api/v1/main/flows/export/by-query"`, so the request did reach the webserver. Report status/route/resolved URL, drop the guess about why.

Root cause is still open. At v2.0.0 that route exists and the handler has no 404 path, it returns 200 with a possibly-empty ZIP. Points at a mismatch in `kestra-api-client:2.0.0` (pinned separately at build.gradle:64). Please keep #19181 open.

Minor: `ApiErrors` could just be `StringUtils.abbreviate(e.getMessage(), 500)`, commons-lang3 is already on the classpath via core.

src/main/java/io/kestra/plugin/git/PushFlows.java
jymaire: yes, we are adding more technical information so users may have a hint of what happened

jymaire:
Thanks @Malaydewangan09 — addressed both merge-blocking asks plus the minor one.

**1. Drop the `kestraUrl` guess (namespace-404 message).** Agreed. The reporter's problem+json body (`"instance": "/api/v1/main/flows/export/by-query"`) proves the request reached the webserver, so blaming `kestraUrl` was unfounded — and doubly so for *this* pre-check, since OSS `GET /namespaces/{id}` doesn't 404 for a missing namespace at all. The message now reports facts only: HTTP status, the resolved route, and the resolved `kestraUrl` (with the fallback-chain note when unset):

> The Kestra API returned HTTP 404 for GET /api/v1/{tenant}/namespaces/{namespace} (resolved `kestraUrl`: …). Verify the namespace exists and that `kestraUrl` resolves to the Kestra webserver API.

`NamespaceSyncTest` now asserts status + route + resolved URL instead of the old cause wording.

**2. Root cause still open — #19181 kept open.** Done. Agreed this PR is only the diagnostics slice: the reporter's `flows/export/by-query` 404 can't come from the v2.0.0 handler (no 404 path there — it returns 200 + a possibly-empty ZIP), so it points at the `kestra-api-client` version rather than a plugin bug. One correction on the pointer, though — the client isn't pinned at `build.gradle:64` on this branch anymore; it now resolves transitively via `plugin-git-lib:1.0.2` (`build.gradle:59`). So the version-mismatch investigation belongs in the lib, and I'll track it under #19181.

**3. `ApiErrors` → `StringUtils.abbreviate`.** Done — deleted the helper and inlined `StringUtils.abbreviate(StringUtils.defaultString(e.getMessage()), 500)` at the four call sites (commons-lang3 is already on the classpath via core, as you noted). Kept `defaultString(...)` so a null SDK message stays `""` rather than printing `null`.

Malaydewangan09 (approved):

PR #348, opened 2026-09-15
fix(sync): create missing namespace in SyncNamespaceFiles (#338)

### LINKED ISSUE #338: `SyncNamespaceFiles` does not create the target namespace if it doesn't exist
## Summary

Running `io.kestra.plugin.git.SyncNamespaceFiles` against a namespace that does not yet exist syncs the files to the instance, but the namespace is not created and the synced files are not visible anywhere in the UI. Files only become visible once the namespace is created manually afterwards.

## Environment

- Kestra: `2.0.0-rc14`
- `plugin-git`: `3.1.0`

## Current behavior

1. Run a flow using `SyncNamespaceFiles` targeting a namespace that does not exist yet.
2. Files are synced to the instance.
3. The namespace and its files are not visible anywhere in the UI.
4. Manually create the namespace afterwards.
5. The previously synced files now appear under it.

## Expected behavior

If a target namespace does not exist, `SyncNamespaceFiles` should create it automatically, consistent with the behavior a user would expect (and initially assumed was already the case) rather than silently syncing files into a namespace that isn't visible or usable until created by hand.

### Environment

- Kestra Version: develop

jymaire:
Closes #338.

## Problem
`SyncNamespaceFiles` writes files straight into internal storage via `runContext.storage().namespace(...)`. When the target namespace doesn't exist yet, the files land on the instance but the namespace is never **registered** as an entity, so it stays invisible in the UI until someone creates it by hand.

## Scope — this is an **EE-scoped** fix, with graceful degradation on OSS
QA (browser-based, both editions) established that **namespace creation is an Enterprise Edition capability**:

- **EE** exposes a namespace-create API (`POST /namespaces`) and manages namespaces as first-class entities. ✅ **Fixed here** — the task now creates the missing namespace and the files are immediately visible.
- **OSS** has **no** namespace-create API (`POST /namespaces` → 404) and no "create namespace" UI — OSS namespaces are purely implicit/derived. There is nothing the plugin can call, so this cannot be fully fixed in the plugin on OSS. The task now **degrades gracefully**: it attempts creation best-effort, and on failure logs a clear warning and lets the sync continue (never hard-fails). Files still sync to storage exactly as before.
- **A complete OSS fix must happen in Kestra core** (e.g. having `Namespace#putFile` register the namespace on write). That's out of scope for this repo — noted here so it isn't lost.

## Implementation
- Ensures the target namespace exists (best-effort, skipped in `dryRun`) before the first write of a run. Logic lives entirely in the concrete `SyncNamespaceFiles` task — **no changes to the shared `plugin-git-lib` kernel**.
- **create-and-swallow-conflict**, not GET-then-create: `GET namespaces/{id}` can't be used as an existence check because it returns `200` with a synthetic body for a never-persisted namespace instead of `404`. We call create directly and swallow the "already exists" response.
- EE reports an already-existing namespace as a **422** validation error (`namespace.id … already exists`), **not 409** — so the conflict check matches on that specific error `path`/`detail` (409 also handled for safety). A genuine validation error (e.g. an invalid id) still surfaces as a warning rather than being silently swallowed.
- No-auth default path is covered by the shared kernel's `auth.auto` mechanism (same as `NamespaceSync`/`TenantSync`/`PushFlows`): with instance-level `kestra.tasks.sdk.authentication.*`, creation happens with zero flow-level credentials. When nothing resolves, warn-and-continue. Both the class-level and `namespace` property `@Schema` now document this.

## QA evidence
Full browser QA across three rounds (OSS KO → EE PASS → idempotency verified). Details posted on #338; screenshot gallery: https://claude.ai/artifact/XuazRhSvW5gUWGE3wWri7t

| Environment | Missing ns → created? | Idempotent re-sync |
|---|---|---|
| **EE** (with auth / instance sdk auth) | ✅ created, visible in UI | ✅ clean, 0 warnings |
| **OSS** | ⚠️ not creatable (no API) — graceful warn-and-continue | ⚠️ same |

## Tests
- `SyncNamespaceFilesTest`: added coverage for missing→created, already-exists (422/409) swallowed with no warning, genuine-422 still warns, dryRun skips, and unreachable/401 warn-and-continue. `NamespaceLifecycleMockServer` models the real 2.x behavior (synthetic-200 GET, 422 on already-exists).
- `SyncNamespaceFilesContainerTest` added (deferred to live-container CI).
- Note: the GitHub-remote unit tests fail in the sandbox for a pre-existing environmental reason (no `GH_PERSONAL_TOKEN` → JGit), unrelated to this change.


Malaydewangan09 (changes requested):
@jymaire no new property?

Edit: my bad its different issue.

src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
```diff
@@ -149,11 +152,76 @@ public class SyncNamespaceFiles extends AbstractSyncTask<NamespaceFile, SyncName
...
+
+    // A 422 validation error is only a benign "already exists" conflict when the offending field is
+    // namespace.id and says so; any other 422 (e.g. an invalid id) must still surface as a failure.
+    private static boolean isNamespaceAlreadyExistsError(ApiException e) {
+        if (e.getCode() != 422 || e.getResponseBody() == null) {
+            return false;
+        }
+        try {
+            for (JsonNode error : JacksonMapper.ofJson().readTree(e.getResponseBody()).path("errors")) {
+                if ("namespace.id".equals(error.path("path").asText()) && error.path("detail").asText("").contains("already exists")) {
```
Malaydewangan09: No better way to find if already exists, rather than a hardcoded message?

Malaydewangan09 (approved):

src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
jymaire: Good catch — done in be7e5af. There's unfortunately no machine-readable discriminator to key on: the EE 422 already-exists body carries only `path`/`pointer`/`detail` per error, no error `code`:

```json
{"status":422,"detail":"namespace.id: Namespace id already exists",
 "errors":[{"detail":"Namespace id already exists","pointer":"/namespace/id","path":"namespace.id"}]}
```

So instead of matching the human message, the helper now keys on the structural field alone — any 422 whose error targets `namespace.id` (via `path` or `pointer`) is treated as a benign already-exists and lets the sync continue. A genuinely invalid id no longer needs distinguishing here: it still hard-fails downstream on the actual file write. No more hardcoded message string.

PR #350, opened 2026-09-15
feat(namespace-files): add namespaceDirectory to scope Sync/Push to a folder

### LINKED ISSUE #349: `SyncNamespaceFiles`: add `namespaceDirectory` to sync into a namespace subfolder
## Problem

`SyncNamespaceFiles` always writes to the namespace root. `gitDirectory` only picks which subtree of the repo to read, it has no effect on where files land in Kestra. A repo whose content sits at the root (`gitDirectory: "."`) can only be dumped into the namespace root, mixed in with whatever is already there.

## Proposal

Add `namespaceDirectory` (`Property<String>`, default `/`) to `SyncNamespaceFiles`, prefixing every destination path. Add the same property to `PushNamespaceFiles` (scopes the exported subtree, strips the prefix) so push/sync stays reversible. The name mirrors `gitDirectory`. `io.kestra.plugin.core.namespace.UploadFiles.destination` already has these semantics in core.

```yaml
- id: sync
  type: io.kestra.plugin.git.SyncNamespaceFiles
  url: https://github.com/company/shared-scripts
  branch: main
  username: git_username
  gitDirectory: "."
  namespace: company.ops
  namespaceDirectory: /shared-scripts
  delete: true
```

Repo root `helpers/parse.py` lands at `company.ops:/shared-scripts/helpers/parse.py`.

## Notes

- `resolveTarget` (`SyncNamespaceFiles.java:349`) is the single place the destination is decided. Prefixing there covers `writeResource`, `simulateResourceWrite` and the `dryRun` diff.
- `delete: true` must be scoped to the prefix. `fetchResources` (`SyncNamespaceFiles.java:319`) returns the whole namespace and `AbstractSyncTask.java:254` deletes anything not present in Git, so without a filter the first sync wipes everything outside the folder. This is the part that needs a test.
- With `includeChildNamespaces: true`, apply the prefix after child-namespace routing.
- Default `/` keeps existing flows unchanged.

Related: #46, #49 (source-side path control).

jymaire:
## Summary

- Add `namespaceDirectory` (`Property<String>`, default `/`) to `SyncNamespaceFiles` and `PushNamespaceFiles`, mirroring `gitDirectory` on the Kestra namespace side.
- `SyncNamespaceFiles`: prefixes every destination path with the (normalized) prefix in `resolveTarget`, applied *after* `includeChildNamespaces` routing so routing still wins; scopes `fetchResources` (and therefore `delete: true`) to Namespace Files already under the prefix, so the first sync never wipes content elsewhere in the namespace.
- `PushNamespaceFiles`: filters `instanceResourcesContentByPath` to files under the prefix and strips it before resolving the Git destination path, so a repo-root `foo.py` round-trips with `namespaceDirectory: /shared-scripts`. The `files` glob keeps matching the full namespace-relative path, independently of the prefix.
- Normalization (duplicated per task, matching the existing `gitDirectory`-per-task pattern in this repo): `/`, `""`, `null` and a trailing slash all collapse to "no prefix"; a leading slash is optional on input; any `..` segment is rejected with an actionable `IllegalArgumentException`.
- Default `/` keeps every existing flow unchanged (regression-tested).

## Test plan

- [x] `rtk test ./gradlew test` passes for the new `namespaceDirectory_*` tests in `SyncNamespaceFilesTest` (default no-op, prefixing, dry-run `kestraPath`, delete scoping leaving out-of-prefix files untouched, delete on an empty Git subtree, leading/trailing-slash acceptance, `..` rejection) and `PushNamespaceFilesTest` (prefix scoping + stripping, `files` glob independence, sync→push round trip)
- [x] The 3 new `PushNamespaceFilesTest` cases talk to `repositoryUrl`/`pat` (GitHub, matching this file's existing convention) and could not be executed in this sandbox (no `GH_PERSONAL_TOKEN` — every pre-existing test using the same fixture fails identically here); the same logic was independently smoke-tested end-to-end (push + strip + sync-back round trip) against the local Gitea instance and passed before being folded into the final tests

closes: https://github.com/kestra-io/plugin-git/issues/349

---
*[View as Artifact](https://claude.ai/artifact/SKMD5HEnTpcipGwZ83LrXy)*


jymaire:
## QA Report — `namespaceDirectory` (#349)

**Edition**: EE | **Docker tag**: `develop` (2.0.1)
**Instance**: `http://localhost:8080` (container `kestra-8uuc0`, left running for manual review — `docker rm -f kestra-8uuc0` to stop)
**Screenshots**: https://claude.ai/artifact/PDDGzmENz6SaX3ecjbdEV6
**Git source**: [`jymaire/qa-ns-dir-349`](https://github.com/jymaire/qa-ns-dir-349) — public throwaway repo, root `README.md` + `helpers/parse.py` + `helpers/util.py`

Scope approved by requester: **Sync-only (flows 1–3)**; the Push round-trip was intentionally deferred (needs a PAT secret).

### Summary

| # | Flow | Result | What it proves |
|---|------|--------|----------------|
| 1 | `sync_prefix` | ✅ SUCCESS | Files land under `namespaceDirectory` |
| 2 | `sync_delete_scoped` | ✅ SUCCESS | **`delete: true` is scoped to the prefix** — the #349 safety guarantee |
| 3 | `sync_default_noop` | ✅ SUCCESS | Non-regression: default (no prefix) unchanged |

---

### Flow 1: sync_prefix (✅ SUCCESS)

```yaml
id: sync_prefix
namespace: qa.prefix
tasks:
  - id: sync
    type: io.kestra.plugin.git.SyncNamespaceFiles
    url: https://github.com/jymaire/qa-ns-dir-349
    branch: main
    gitDirectory: "."
    namespace: qa.prefix
    namespaceDirectory: /shared-scripts
```
</details>

**Gantt** ([screenshot](https://claude.ai/artifact/PDDGzmENz6SaX3ecjbdEV6#sync_prefix-gantt))

| Task | Status |
|------|--------|
| `sync` | SUCCESS |

**Result** — verified via the namespace-files API:
- `qa.prefix` root contains **only** `shared-scripts/` (nothing dumped at root).
- `qa.prefix:/shared-scripts/README.md`, `/shared-scripts/helpers/parse.py`, `/shared-scripts/helpers/util.py` all present.

Repo-root content correctly prefixed into the target subfolder. ([outputs screenshot](https://claude.ai/artifact/PDDGzmENz6SaX3ecjbdEV6#sync_prefix-outputs))

---

### Flow 2: sync_delete_scoped (✅ SUCCESS — key test)

Pre-seeded before the run: `qa.delete:/keep-me.txt` (**outside** the prefix) and `qa.delete:/shared-scripts/stale.txt` (**inside** the prefix, absent from Git).

```yaml
id: sync_delete_scoped
namespace: qa.delete
tasks:
  - id: sync
    type: io.kestra.plugin.git.SyncNamespaceFiles
    url: https://github.com/jymaire/qa-ns-dir-349
    branch: main
    gitDirectory: "."
    namespace: qa.delete
    namespaceDirectory: /shared-scripts
    delete: true
```
</details>

**Gantt** ([screenshot](https://claude.ai/artifact/PDDGzmENz6SaX3ecjbdEV6#sync_delete_scoped-gantt)) — task `sync` SUCCESS.

**Result** — verified via API and the Namespace Files UI ([screenshot](https://claude.ai/artifact/PDDGzmENz6SaX3ecjbdEV6#sync_delete_scoped-nsfiles)):

| File | Location vs prefix | Before | After | Expected |
|------|--------------------|--------|-------|----------|
| `keep-me.txt` | outside | present | **present** ✅ | survives |
| `shared-scripts/stale.txt` | inside, not in Git | present | **deleted** ✅ | removed |
| `shared-scripts/README.md`, `helpers/*` | inside, from Git | – | present ✅ | synced |

This is the guarantee #349 calls out: **a first `delete: true` sync into a subfolder never wipes content elsewhere in the namespace**, while delete still works within the prefix.

---

### Flow 3: sync_default_noop (✅ SUCCESS — non-regression)

```yaml
id: sync_default_noop
namespace: qa.noop
tasks:
  - id: sync
    type: io.kestra.plugin.git.SyncNamespaceFiles
    url: https://github.com/jymaire/qa-ns-dir-349
    branch: main
    gitDirectory: "."
    namespace: qa.noop
```
</details>

**Gantt** ([screenshot](https://claude.ai/artifact/PDDGzmENz6SaX3ecjbdEV6#sync_default_noop-gantt)) — SUCCESS.

**Result** — with no `namespaceDirectory`, files land at the namespace **root** (`/README.md`, `/helpers/parse.py`, `/helpers/util.py`) exactly as before. Default `/` is a true no-op. ✅

---

### Observation (not a blocker)

With `gitDirectory: "."` the repo's `.git` directory is also synced (appears as `shared-scripts/.git` in flows 1–2, `/.git` in flow 3). This is **pre-existing** `gitDirectory: "."` behavior — identical across all three flows and independent of the prefix logic — but #349's headline use case (repo root → subfolder) makes it more visible. Worth a follow-up to exclude `.git` from the synced tree, separate from this PR.

### Timeouts / skips

- None. All three executions completed in a few seconds.
- Push (`PushNamespaceFiles`) round-trip deferred per approved scope — covered by unit tests in this PR.

---
*Automated browser QA via `/kestra-plugins:doing-qa`.*

jymaire:
## Docs impact assessment

**docs_required:** likely (external docs only — no in-PR action needed)

**user_impact:** Two new optional plugin properties (`SyncNamespaceFiles.namespaceDirectory`, `PushNamespaceFiles.namespaceDirectory`) let users sync/push into a namespace subfolder, with `delete: true` scoped to the prefix. Default `/` preserves existing behavior.

**affected_audience:** Kestra users adopting Git-based version control of namespace files.

**In-PR docs surface — already complete, no changes requested:**
- Both tasks carry detailed `@Schema` title + description explaining the pairing with `gitDirectory` and the no-prefix/delete-scoping semantics — these generate the plugin reference automatically.
- Both tasks add an `@Example` demonstrating the repo-root pattern (`gitDirectory: "."` + `namespaceDirectory: /shared-scripts`).

**Follow-up (separate `kestra-io/docs` task, not this PR):** the narrative how-to guides would benefit from showing the new property and the repo-root → subfolder pattern:
- `src/contents/docs/15.how-to-guides/syncnamespacefiles/index.md`
- `src/contents/docs/15.how-to-guides/pushnamespacefiles/index.md`
- `src/contents/docs/version-control-cicd/04.git`

**gaps_or_unknowns:** none — the in-repo docs surface was verified complete against the diff.

_Assessed via `/kestra-docs:impact-assessor`. No follow-up commit needed on this PR; docs work tracked separately in kestra-io/docs._

src/main/java/io/kestra/plugin/git/PushNamespaceFiles.java
```diff
@@ -179,6 +219,7 @@ protected Map<Path, Supplier<InputStream>> instanceResourcesContentByPath(RunCon
 
         String renderedNamespace = runContext.render(this.namespace).as(String.class).orElse(null);
         Predicate<Path> matcher = (globs != null) ? PathMatcherPredicate.matches(globs) : (path -> true);
+        String prefix = normalizeNamespaceDirectory(runContext.render(this.namespaceDirectory).as(String.class).orElse("/"));
```
jymaire: 🔴 [Correctness/Data-loss] Push `delete` (default true) is not scoped to `namespaceDirectory`

Problem: `instanceResourcesContentByPath` filters the returned `contentByPath` map to only files under the `namespaceDirectory` prefix, but `AbstractPushTask.deleteOutdatedResources` (plugin-git-lib, unmodified) walks the *entire* `gitDirectory` on disk and stages `git rm` for every file not present in `contentByPath` — filtered only by the independent `globs`/`files` matcher, which `namespaceDirectory` never feeds into. Since `delete` defaults to `true`, adding `namespaceDirectory` to an existing `PushNamespaceFiles` task (e.g. one that has been pushing a whole namespace to `_files` for months) will, on the very next run, delete and push the removal of every previously-pushed file outside the new prefix — the exact "first scoped run wipes everything else" failure mode the issue explicitly requires to be prevented (and which `SyncNamespaceFiles.fetchResources` correctly avoids). None of the three new `namespaceDirectory_*` tests seed a pre-existing file directly in the Git repo before pushing with default `delete=true`, so this regression is completely undetected.

Fix: Since `plugin-git-lib` cannot be modified, scope the deletion from inside this class instead of relying on the library's directory walk being prefix-aware. One approach: before returning `filesMap`, also walk the already-checked-out `baseDirectory` for existing files that fall *outside* the computed prefix and re-add them to `filesMap` unchanged (reading their own on-disk content back), so they land in the library's `keep` set and survive `deleteOutdatedResources`. At minimum, until a safe mechanism exists, this interaction must be covered by a test that seeds an out-of-prefix file directly in the Git branch and asserts it survives a scoped push with default `delete=true`.

src/test/java/io/kestra/plugin/git/PushNamespaceFilesTest.java
```diff
@@ -619,6 +620,198 @@ void kestraIgnore_IgnoresPushAndSkipsDeletion() throws Exception {
         }
     }
 
+    @Test
+    void namespaceDirectory_ScopesAndStripsPushedPaths() throws Exception {
```
jymaire, line 624: 🟠 [Tests] Missing coverage: namespaceDirectory + includeChildNamespaces, and delete-safety against pre-existing Git content

Problem: All three new tests only seed content in Kestra namespace storage, never in the Git branch itself, and none set `includeChildNamespaces: true`. This leaves two real interactions unverified: (1) whether `namespaceDirectory` scoping is applied consistently per child namespace on push, and (2) whether `delete` (default true) leaves Git files outside the prefix untouched (see the Block finding on `instanceResourcesContentByPath`).

Fix: Add a test that seeds a file directly in the cloned Git working tree outside `namespaceDirectory` before running a scoped push with default `delete=true`, asserting it still exists after the push/clone-verify — mirroring `SyncNamespaceFilesTest.namespaceDirectory_Delete_LeavesFilesOutsidePrefixUntouched`. Add a second test combining `namespaceDirectory` with `includeChildNamespaces: true`.

src/test/java/io/kestra/plugin/git/SyncNamespaceFilesTest.java
```diff
@@ -297,6 +299,179 @@ void syncWithDotGitInDirectoryName_ShouldSyncFiles() throws Exception {
...
+        List<Map<String, String>> diffs = readDiffs(runContext, output.diffFileUri());
+        assertThat(
+            diffs.stream().map(diff -> diff.get("kestraPath")).toList(),
+            hasItem("/my/namespace/_files/shared-scripts/foo.py")
+        );
+        assertThat(runContext.storage().namespace(NAMESPACE).exists(Path.of("shared-scripts/foo.py")), is(false));
+    }
+
+    @Test
+    void namespaceDirectory_Delete_LeavesFilesOutsidePrefixUntouched() throws Exception {
```
jymaire, line 358: 🟠 [Tests] Missing coverage: namespaceDirectory + includeChildNamespaces combined

Problem: The delete-scoping tests here only cover the primary namespace. `isUnderNamespaceDirectory` is applied identically to child-namespace resources in `fetchResources`, but no test exercises `includeChildNamespaces: true` together with `namespaceDirectory` and `delete: true`, so cross-namespace delete-scoping (explicitly called out as needing a test in issue #349) is unverified for the child-namespace path.

Fix: Add a test with `includeChildNamespaces: true` and `namespaceDirectory` set, seeding both an in-prefix and an out-of-prefix file in a child namespace, and assert only the in-prefix one is affected by `delete: true`.

src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
```diff
@@ -166,6 +209,44 @@ private List<String> childNamespaces(RunContext runContext, String renderedNames
...
+                );
+            } catch (IllegalVariableEvaluationException e) {
+                throw new IOException(e);
+            }
+        }
+        return this.resolvedNamespaceDirectory;
+    }
+
+    // "/", "" and null all normalize to "" (no prefix); a leading slash is optional on input and forced on output
+    private static String normalizeNamespaceDirectory(String rendered) {
```
jymaire: 🟠 [Guidelines/DRY] normalizeNamespaceDirectory duplicated verbatim across two task classes

Problem: This exact method (implementation, and even the same leading comment) is copy-pasted into `PushNamespaceFiles.java`. Per the DRY guideline, duplicated logic across tasks in the same repository must be extracted into a shared utility (this cannot go into `plugin-git-lib` per AGENTS.md, but a small local helper class in this repo is fine and expected).

Also note as a minor edge case: segments are filtered only for emptiness, not for a literal `.` segment — `namespaceDirectory: "./shared-scripts"` normalizes to `/./shared-scripts` instead of `/shared-scripts` or being rejected, silently producing a prefix that will never match any real path (so it behaves as "match nothing" rather than failing loudly or normalizing as a user would expect).

Fix: Extract `normalizeNamespaceDirectory` (and the sync/strip predicates) into a single package-private utility (e.g. `io.kestra.plugin.git.NamespaceDirectories`) used by both `SyncNamespaceFiles` and `PushNamespaceFiles`. While there, also filter out `.` segments alongside empty ones.

src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
```diff
@@ -149,6 +188,10 @@ public class SyncNamespaceFiles extends AbstractSyncTask<NamespaceFile, SyncName
     @ToString.Exclude
     private transient List<String> resolvedChildNamespaces;
 
+    @Getter(AccessLevel.NONE)
+    @ToString.Exclude
+    private transient String resolvedNamespaceDirectory;
```
jymaire: 🟡 [Performance/Thread-safety] Lazily-cached prefix on a potentially shared Task instance

Problem: `resolvedNamespaceDirectory` (like the pre-existing `resolvedChildNamespaces`) is a plain, unsynchronized `transient` field lazily populated with a null-check. If the same parsed `SyncNamespaceFiles` instance is invoked concurrently for two different executions of the same flow revision (e.g. `namespaceDirectory` driven by `{{ inputs.folder }}` in a flow run concurrently for two different inputs), the second execution can silently read the first execution's cached, wrong prefix instead of rendering its own.

Fix: Not a new pattern introduced by this PR, but this doubles down on it. Consider computing the prefix once in a local variable and threading it through the template methods instead of caching on the instance, or confirm with the framework owners that `Task` instances are never shared across concurrent executions of the same flow (in which case this is fine as-is and can be left alone).

src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
```diff
@@ -346,26 +434,29 @@ protected URI toUri(String renderedNamespace, NamespaceFile resource) {
...
-        if (slash < 0) {
-            return new Target(renderedNamespace, uri);
-        }
-        String first = raw.substring(0, slash);
-        if (!isDescendant(renderedNamespace, first) || !childNamespaces(runContext, renderedNamespace).contains(first)) {
-            return new Target(renderedNamespace, uri);
-        }
-        return new Target(first, URI.create(raw.substring(slash)));
+        String prefix = this.namespaceDirectoryPrefix(runContext);
+        return prefix.isEmpty() ? routed : new Target(routed.namespace(), URI.create(prefix + routed.uri()));
```
jymaire: 🟡 [Error handling] Raw URI.create can throw an opaque exception for valid-looking namespaceDirectory values

Problem: `prefix` is a raw, non-URI-encoded string built from user input segments. If `namespaceDirectory` contains a character that is illegal in a URI path (space, `#`, `%`, `?`, ...), `URI.create(prefix + routed.uri())` throws a raw `java.net.URISyntaxException`-derived `IllegalArgumentException` with a low-level "Illegal character" message, not an actionable Kestra error — violating the "Error Messages" guideline (no raw SDK/internal error type as the top-level message).

Fix: Percent-encode each normalized segment before building the prefix (mirroring how `NamespaceFile` paths are normally constructed), or catch `IllegalArgumentException` here and rethrow with a clear message naming the offending `namespaceDirectory` value.

jymaire (commented):
## Kestra Plugin Code Review

### Business Requirements — partial
Core ask from #349 is implemented: `namespaceDirectory` prefixes Sync destinations, scopes/strips on Push, defaults to `/` (no-op), and the Sync-side delete-scoping the issue explicitly calls out as "the part that needs a test" is correctly implemented and tested. However, the issue's own stated goal that "push/sync stays reversible" is undermined by a real data-loss regression on the Push side (see Block finding below) that isn't covered by any test.

- 🟠 DRY: `normalizeNamespaceDirectory` duplicated verbatim between `SyncNamespaceFiles` and `PushNamespaceFiles` (see inline).
- 🟡 Doc: `src/main/resources/doc/io.kestra.plugin.git.md`'s Tasks section documents `SyncNamespaceFiles`/`PushNamespaceFiles` but was not updated to mention `namespaceDirectory`, per "Plugin How-To Doc" guideline (update the matching section when a property is added). Low severity given the doc is already terse and doesn't enumerate `gitDirectory`/`includeChildNamespaces` either.
- 🟡 Thread-safety: see inline on `resolvedNamespaceDirectory`.

### Security (OWASP Top 10:2025 + KPS) — 1 finding
- 🟡 A10/Error handling: unescaped `namespaceDirectory` segments can make `URI.create` throw a raw, low-level `IllegalArgumentException` instead of an actionable Kestra error (see inline). No traversal/injection issue — `..` is explicitly and correctly rejected, and the storage layer is namespace/tenant-scoped throughout.

Prefix/child-namespace values are rendered and normalized once per run and reused (caching is the only concern, filed under thread-safety, not perf). No unbounded allocation, no per-iteration regex/string-concat hot paths introduced.

### Additional — Test coverage — 2 findings
- 🟠 Neither `SyncNamespaceFilesTest` nor `PushNamespaceFilesTest` combines `namespaceDirectory` with `includeChildNamespaces`, so delete-scoping across child namespaces (explicitly called out in the review brief as a required check) is unverified in either direction.

**Verdict:** BLOCK

src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
jymaire: Fixed in 742e9fd: removed the resolvedNamespaceDirectory instance cache; the prefix is now recomputed per run instead of memoized on the (potentially concurrently reused) Task instance.

src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
jymaire: Fixed in 742e9fd: resolveTarget now wraps URI.create and rethrows IllegalArgumentException as an IOException naming the offending namespaceDirectory value, instead of surfacing the raw 'Illegal character' message.

src/test/java/io/kestra/plugin/git/SyncNamespaceFilesTest.java
jymaire: Fixed in 742e9fd: added namespaceDirectory_Delete_ChildNamespace_ScopesToPrefix — includeChildNamespaces:true + namespaceDirectory + delete:true, seeding in-prefix and out-of-prefix files in a child namespace, asserting only the in-prefix one is affected.

src/main/java/io/kestra/plugin/git/PushNamespaceFiles.java
jymaire: Fixed in 742e9fd: PushNamespaceFiles now re-adds out-of-prefix files already checked out in gitDirectory back into the returned map (reading their on-disk content) so they land in deleteOutdatedResources' keep set — a first scoped push with default delete:true no longer wipes files outside the prefix. Covered by namespaceDirectory_Delete_PreservesFilesOutsidePrefix.

src/test/java/io/kestra/plugin/git/PushNamespaceFilesTest.java
jymaire: Fixed in 742e9fd: added namespaceDirectory_Delete_PreservesFilesOutsidePrefix (seeds an out-of-prefix file directly in the Git branch, asserts it survives a scoped push with default delete:true) and namespaceDirectory_IncludeChildNamespaces_ScopesEachNamespace.

src/main/java/io/kestra/plugin/git/SyncNamespaceFiles.java
jymaire: Fixed in 742e9fd: extracted normalize/isUnderPrefix/stripPrefix into a package-private io.kestra.plugin.git.NamespaceDirectories helper (local to this repo, not plugin-git-lib) shared by both tasks. normalize now also drops literal '.' segments, so ./shared-scripts normalizes to /shared-scripts.

jymaire:
## QA — finding #1 (🔴 scoped-push data loss)

Verified the fix with a **reproduce-then-fix** run of the network-free regression test `PushNamespaceFilesTest#namespaceDirectory_Delete_PreservesFilesOutsidePrefix`.

The test seeds an out-of-prefix file already on the branch (`my-files/existing/outside.txt`), then runs a scoped push with `namespaceDirectory: /shared-scripts` and the default `delete: true`.

| Run | Code state | Result |
|-----|-----------|--------|
| **1 — bug reproduced** | preserve logic disabled | ❌ FAIL — `AssertionError: Expected: is <true> but: was <false>` on `preserved.exists()`. The out-of-prefix file is deleted → data loss. |
| **2 — fix applied** (`742e9fd`) | preserve logic active | ✅ PASS — `existing/outside.txt` survives with original content, and the newly scoped `keep.txt` is pushed. |

Run 1 reproduces exactly the reported symptom: a first scoped push silently wipes everything outside the new prefix. Run 2 confirms the fix re-adds out-of-prefix files to the library keep set so `AbstractPushTask#deleteOutdatedResources` leaves them intact.

Build state: `compileJava` / `compileTestJava` clean, all 4 new tests pass. The remaining pre-existing failures in the suite are infra-dependent (need `GH_PERSONAL_TOKEN` / a running Gitea) and touch none of the changed code.

src/main/java/io/kestra/plugin/git/NamespaceDirectories.java
```diff
@@ -0,0 +1,70 @@
...
+     * dropped like empty ones so they don't silently produce a prefix that matches nothing.
+     */
+    static String normalize(String rendered) {
+        if (rendered == null) {
+            return "";
+        }
+        List<String> segments = Arrays.stream(rendered.trim().split("/"))
+            .filter(segment -> !segment.isEmpty() && !".".equals(segment))
+            .toList();
+        if (segments.contains("..")) {
```
fdelbrayelle: 🟠 [Security] Path-traversal guard bypassable via percent-encoding
Problem: normalize() rejects a segment only when it is literally ".." after splitting the raw string on "/". A namespaceDirectory value containing a percent-encoded segment (e.g. `%2e%2e/etc`) passes this check untouched (segment == "%2e%2e", not ".."). The resulting prefix is later concatenated into a URI in SyncNamespaceFiles.resolveTarget via `URI.create(prefix + routed.uri())`, and `uri.getPath()` decodes percent-encoding, turning `%2e%2e` back into `..` — so the destination Path used by writeResource/simulateResourceWrite can escape the intended namespace subtree. namespaceDirectory is a Property<String>, so it can be driven by flow inputs (e.g. a webhook body) in some flow designs, making this attacker-reachable in principle. The advertised guarantee in the PR description ("any .. segment is rejected") does not hold for encoded input.
Fix: Validate after resolving to a real Path/URI rather than on the raw string — e.g. build the candidate URI first, call .normalize(), and assert the decoded path has no ".." segment and still starts with the expected root, instead of (or in addition to) the current raw split-on-"/" check.

src/main/java/io/kestra/plugin/git/PushNamespaceFiles.java
```diff
@@ -179,29 +225,99 @@ protected Map<Path, Supplier<InputStream>> instanceResourcesContentByPath(RunCon
...
+            if (filesMap.containsKey(file) || excludedDirectories.stream().anyMatch(file::startsWith)) {
+                continue;
+            }
+            Path relative = directory.relativize(file);
+            if (isInsideGitMetadataDirectory(relative)) {
+                continue;
+            }
+            String candidatePath = "/" + relative.toString().replace('\\', '/');
+            if (!NamespaceDirectories.isUnderPrefix(candidatePath, prefix)) {
+                byte[] content = Files.readAllBytes(file);
```
fdelbrayelle: 🟠 [Performance] Eager full-file buffering instead of lazy streaming
Problem: For every already-checked-out file outside the new prefix, this reads the entire file into a byte[] immediately (Files.readAllBytes) and stores it in filesMap, rather than deferring the read like the lazy Supplier<InputStream> used a few lines above (throwSupplier(() -> storage.getFileContent(...))) for in-prefix files. On a first scoped push against a repository that already has many or large pre-existing files outside the chosen namespaceDirectory, this can buffer a large amount of file content in memory at once before anything is committed.
Fix: Keep this lazy too, e.g. filesMap.put(file, () -> { try { return Files.newInputStream(file); } catch (IOException e) { throw new UncheckedIOException(e); } }); — matches the existing lazy convention in this same method and avoids materializing content unnecessarily.

src/main/java/io/kestra/plugin/git/PushNamespaceFiles.java
```diff
@@ -139,6 +169,22 @@ public class PushNamespaceFiles extends AbstractPushTask<PushNamespaceFiles.Outp
...
+            under this path and the prefix is stripped from the destination path in Git, so a Namespace File at \
+            `/shared-scripts/foo.py` is pushed as `foo.py` inside `gitDirectory` instead of \
+            `shared-scripts/foo.py`. Defaults to `/` (namespace root, i.e. no prefix — existing flows are \
+            unaffected). The `files` glob still matches against the full namespace-relative path, independently \
+            of this prefix. Also scopes `delete` (default true) to files already under this prefix in \
+            `gitDirectory`, so files previously pushed from outside this prefix are left untouched even on the \
+            very first scoped push."""
+    )
+    @Builder.Default
+    @PluginProperty(group = "destination")
```
fdelbrayelle, line 185: 🟢 [Guidelines] Property group placement
Problem: namespaceDirectory is grouped 'destination', but it scopes which Namespace Files are read from the source namespace (same role as namespace and includeChildNamespaces, both grouped 'source'), splitting closely related source-scoping properties across two UI sections.
Fix: Consider group = "source" to match namespace/includeChildNamespaces, or leave a short note explaining why it stays paired with gitDirectory under 'destination' (it also affects the Git-side path). Non-blocking.

src/test/java/io/kestra/plugin/git/NamespacesSearchMockServer.java
```diff
@@ -0,0 +1,55 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+import com.sun.net.httpserver.HttpServer;
+
+/**
+ * Minimal in-process stand-in for the Kestra {@code GET /api/v1/{tenant}/namespaces/search} endpoint used only to
+ * exercise {@code includeChildNamespaces} in {@link SyncNamespaceFilesTest} / {@link PushNamespaceFilesTest}.
+ *
+ * <p>
```
fdelbrayelle, line 14: 🟡 [Guidelines] Hand-rolled HTTP mock vs. project test-infra convention
Problem: This introduces a second, bespoke HTTP-mocking mechanism (raw com.sun.net.httpserver.HttpServer) alongside plugin-git-lib's MockKestraApiServer. Guidelines call for Testcontainers first, falling back to WireMock (org.wiremock:wiremock-jetty12) when Testcontainers isn't suitable for API mocking — this repo has neither a WireMock dependency nor this pattern used elsewhere.
Fix: Non-blocking given the narrow, well-documented scope (single endpoint, test-only, properly closed) — but consider migrating to WireMock if more endpoints need mocking in the future, to keep one convention.

fdelbrayelle (changes requested):
## Kestra Plugin Code Review

Implements everything asked in #349: `namespaceDirectory` (`Property<String>`, default `/`) added to both `SyncNamespaceFiles` and `PushNamespaceFiles`; prefixing applied after `includeChildNamespaces` routing in `resolveTarget`; `fetchResources`/`instanceResourcesContentByPath` scoped to the prefix so `delete: true` never wipes content outside it on a first scoped run; `files` glob still matches the full namespace-relative path; default `/` keeps existing flows unchanged. Well covered by new happy-path, dry-run, delete-scoping, and round-trip tests in both `SyncNamespaceFilesTest` and `PushNamespaceFilesTest`. No scope creep — no re-implementation of `plugin-git-lib` plumbing; the new `NamespaceDirectories` helper is correctly scoped as task-local convention per AGENTS.md.

- 🟠 Mandatory package-level how-to doc (`src/main/resources/doc/io.kestra.plugin.git.md`) was not updated. It still only says 'SyncNamespaceFiles to pull namespace files... PushNamespaceFiles for namespace files' with no mention of the new `namespaceDirectory` property or its delete-scoping/prefix-stripping gotchas. Per guidelines, an existing-plugin property addition **must** update the matching section — listed as Mandatory, not optional. Add a short paragraph/bullet under the relevant Tasks description.
- 🟡 See inline comment on `NamespacesSearchMockServer.java` (test infra fallback convention).
- 🟢 See inline comment on `PushNamespaceFiles.java:185` (property group placement nit).

### Security (OWASP Top 10:2025 + KPS) — 1 finding
- 🟠 See inline comment on `NamespaceDirectories.java:34` — the `..` path-traversal guard is bypassable via percent-encoding, undermining the safety claim made in the PR description ('any `..` segment is rejected').
- No secret-handling, HTTP-client, or deserialization issues found elsewhere in the diff. `namespaceDirectory` correctly stays a non-secret `Property<String>`; no new external calls, no new dependencies.

### Performance — 1 finding
- 🟠 See inline comment on `PushNamespaceFiles.java:306` — `preserveFilesOutsidePrefix` eagerly reads every out-of-prefix file fully into memory (`Files.readAllBytes`) instead of lazily streaming, unlike the lazy `Supplier<InputStream>` pattern used a few lines above in the same method for in-prefix files. Risk of high memory use on a first scoped push against a repo with many/large pre-existing files.
- No other memory-leak, unbounded-allocation, or O(n²) issues found; the new prefix filtering (`isUnderPrefix`/`stripPrefix`) is O(1) per file and the new stream filters in `fetchResources` are linear.

**Verdict:** REQUEST CHANGES

jymaire (commented):
Thanks for the thorough review @fdelbrayelle — addressed in 14e21ef.

### 🟠 Security — percent-encoded `..` bypass (`NamespaceDirectories.java`)
Good catch. `normalize()` now validates the prefix **after** a single URI decode (`URI.create(prefix).getPath()`), which mirrors exactly how the prefix is decoded when concatenated in `SyncNamespaceFiles.resolveTarget`. So `/%2e%2e/escape` now decodes to `..` and is rejected just like a literal `../escape`; double-encoded input (`%252e…`) stays a literal segment after one decode, matching the single decode that actually happens downstream, so it can't traverse either. Added `namespaceDirectory_RejectsPercentEncodedPathTraversal` to lock this in.

### 🟠 Performance — eager `readAllBytes` (`PushNamespaceFiles.java`)
Fixed. `preserveFilesOutsidePrefix` now defers with `() -> Files.newInputStream(file)` (wrapped `UncheckedIOException`), matching the lazy `Supplier<InputStream>` convention used for in-prefix files a few lines up — out-of-prefix files are only read when actually committed, so no bulk buffering on a first scoped push.

### 🟠 Docs — package how-to (`io.kestra.plugin.git.md`)
Added a **"Scoping namespace files to a subfolder"** section covering the prefix mapping (both directions), the two gotchas (delete scoping, `files`-glob independence), and the `..` rejection.

### 🟢 Property group placement (`PushNamespaceFiles.java:185`) — kept as `destination`
Deliberately keeping `destination` for symmetry: `namespaceDirectory` is the Kestra-side counterpart of `gitDirectory` and, on push, it also rewrites the Git destination path (prefix stripping), so it belongs paired with `gitDirectory`. Keeping it identical across `SyncNamespaceFiles` (where it genuinely is the destination) and `PushNamespaceFiles` avoids the two tasks disagreeing on the same property's group.

### 🟡 Hand-rolled HTTP mock (`NamespacesSearchMockServer.java`) — kept for now
Agree on the convention direction. Scope here is a single test-only endpoint that's already closed properly, and adding a WireMock dependency for it felt heavier than the problem. Left as-is per your "non-blocking" note; happy to migrate to WireMock the moment a second endpoint needs mocking.

fdelbrayelle (approved):

src/main/java/io/kestra/plugin/git/NamespaceDirectories.java
jymaire: Fixed in 14e21ef: the traversal guard no longer checks the raw string. `NamespaceDirectories.containsParentSegment` now percent-decodes the prefix via `URI.create(prefix).getPath()` (a single decode pass, mirroring how `SyncNamespaceFiles.resolveTarget` later decodes it) and only then splits on `/` and rejects any `..` segment — so `%2e%2e/etc` is caught and rejected, and the "any `..` is rejected" guarantee now holds for encoded input too.

src/main/java/io/kestra/plugin/git/PushNamespaceFiles.java
jymaire: Fixed in 14e21ef: the out-of-prefix preservation branch (`preserveFilesOutsidePrefix`) no longer calls `Files.readAllBytes`. It now stores a lazy `Supplier<InputStream>` (`Files.newInputStream(file)`, wrapped for the checked exception), matching the in-prefix `throwSupplier(() -> storage.getFileContent(...))` convention a few lines above, so nothing is buffered until the library actually consumes it.

src/main/java/io/kestra/plugin/git/PushNamespaceFiles.java
jymaire: Keeping `group = "destination"` intentionally: `namespaceDirectory` scopes the source read, but it *also* rewrites the Git-side destination path (it strips the prefix before resolving the Git path, and scopes `delete` to `gitDirectory` content already under it), so it's paired with `gitDirectory` rather than `namespace`/`includeChildNamespaces`. The schema description now spells out that dual role. Happy to move it to "source" if you feel the source-scoping angle dominates — non-blocking either way.

src/test/java/io/kestra/plugin/git/NamespacesSearchMockServer.java
jymaire: Agreed — kept the hand-rolled `com.sun.net.httpserver.HttpServer` for now given the narrow, test-only, single-endpoint scope (properly closed, no extra dependency). If more endpoints need mocking we'll migrate to WireMock (`org.wiremock:wiremock-jetty12`) to converge on one convention. Noting it as a follow-up rather than blocking this PR.

PR #351, opened 2026-09-16
test(sync-namespace-files): add two-run delete regression coverage for #330

### LINKED ISSUE #330: `SyncNamespaceFiles` fails to delete namespace files removed from Git with `delete: true`
## Summary

When using `SyncNamespaceFiles` with `delete: true`, namespace files that are removed from the Git repository are not deleted from Kestra on the next sync. The file is correctly created on the initial sync, but a subsequent sync after the file is removed from Git leaves the namespace file in place instead of removing it.

## Steps to reproduce

1. Commit a file to the Git repository used by the flow below.
2. Run the flow → the namespace file is created in Kestra.
3. Remove the file from the Git repository.
4. Run the flow again → the namespace file is still present in Kestra. Expected: the file should be removed, since `delete: true` is set.

### Minimal reproducer flow

```yaml
id: git_sync
namespace: system

tasks:
  - id: sync_files
    type: io.kestra.plugin.git.SyncNamespaceFiles
    username: <username>
    url: <repo-url>
    branch: <branch>
    namespace: git
    delete: true
```

### Environment

- Kestra Version: `2.0.0-RC12`
- plugin-git: `3.0.0`

jymaire:
## Summary

- Investigated #330 (`SyncNamespaceFiles` with `delete: true` not deleting a file removed from Git on a subsequent sync, reported on Kestra 2.0.0-RC12 / plugin-git 3.0.0).
- Added a two-run in-process regression test (`sync` -> remove file from a local Git repo -> `sync` again) exercising the exact reported scenario against the shared delete/keep computation in `AbstractSyncTask#run` (plugin-git-lib). **It passes**: the file is correctly removed from the namespace and reported as `DELETED` in the second sync's diff. The delete logic itself is correct in-process, so this does not reproduce the reported behaviour here.
- Given that, the most plausible root cause is environmental — most likely a `gitDirectory` that no longer points at the same path the namespace content was originally synced from (`SyncNamespaceFiles.gitDirectory` defaults to `_files`), which silently no-ops `delete: true` instead of failing loudly.
- Filed a companion defensive fix in the shared kernel: kestra-io/plugin-git-lib#15 adds a `WARN` log when `delete: true` resolves to an empty deletion set on a non-empty namespace, naming the resource count and the rendered `gitDirectory`, so a misconfiguration like this is no longer silent. Not consumed here yet — bumping `plugin-git-lib` to a version containing that fix is deferred until it is reviewed/released (the version currently pinned, `1.0.2`, is already published and this repo's regression test already proves it deletes correctly).

## Test plan

- [x] `rtk test ./gradlew test --tests '*SyncNamespaceFilesTest*'` — the two local-Git-repo tests (`syncWithDotGitInDirectoryName_ShouldSyncFiles`, `secondSyncAfterFileRemovedFromGit_WithDelete_ShouldDeleteFile`) pass; the four tests hitting the real GitHub-backed fixture repo fail in this sandbox with `not authorized` (no `GITHUB_ACCESS_TOKEN`/`kestra.git.pat` configured locally) — pre-existing environmental gap, unrelated to this change.
- [ ] Once kestra-io/plugin-git-lib#15 is reviewed and released, bump `plugin-git-lib` here and re-verify

closes: https://github.com/kestra-io/plugin-git/issues/330


Malaydewangan09 (approved):
Some conflicts to resolve.

PR #353, opened 2026-09-16
fix(sync-flows): surface unresolved flow lookups on SyncFlows.Output

### LINKED ISSUE #336: SyncFlows: surface unresolved flow lookups on Output instead of per-flow WARN logs only
Follow-up from the review of #333 (fix for #327).

### Problem

`SyncFlows.fetchFlowFromApi` now correctly distinguishes a `404` ("flow does not exist") from any other API failure, and logs a `WARN` per flow when a lookup cannot be resolved. The run then continues and falls back to source-text comparison, so the flow is typically reported as `UNCHANGED`.

That is the right behaviour for an isolated failure, but the only signal is a scattered per-flow log line. If the underlying cause is systemic — the calling token lost read access to the single-flow endpoint, or the API is degraded — then *every* flow in the namespace is affected, and a user reading only the diff output sees a clean, entirely `UNCHANGED` result with no indication that the comparison was made on degraded input.

### Suggestion

Track how many lookups went unresolved during a run and surface it on `SyncFlows.Output` — a count, or a boolean flag — so a systemic failure is visible without grepping logs. Worth considering whether a fully-unresolved run should warn more loudly than a partial one.

### Notes

- This adds a key to the task's public output surface, so it needs a docs pass (`revision`-style `@Schema` annotation) and a test alongside it.
- Kept out of #333 to keep that PR scoped to the #327 NPE fix.

Raised by @fdelbrayelle in review of #333.

jymaire:
## Summary

- `SyncFlows.fetchFlowFromApi` distinguishes a 404 ("flow does not exist") from any other API failure, and on a non-404 failure only logs a per-flow WARN and falls back to source-text comparison (typically reporting the flow as UNCHANGED). If the failure is systemic (token lost read access to the single-flow endpoint, or the API is degraded), every flow is affected but the run only ever surfaced a clean all-UNCHANGED diff with no aggregate signal.
- Added `unresolvedFlowLookups` (`Integer`, always populated, 0 on a clean run) to `SyncFlows.Output`, counted via a per-invocation counters holder set in `run()` and cleared in a `finally` block once `output()` has read it. Both the dry-run (`simulateResourceWrite`) and non-dry-run (`writeResource` re-fetch) paths count their lookups. A 404 is a resolved outcome and is never counted as unresolved.
- The counters are held in a `static ThreadLocal`, not a shared instance field: a task instance can be reused across concurrent invocations (e.g. the same templated child task invoked once per parallel-loop iteration, or two overlapping executions of the same flow), and resetting a shared instance field at the top of `run()` is only safe for sequential reuse of the same instance — under real concurrency, one invocation's reset or increments can bleed into another's final count. The `ThreadLocal` gives each invocation its own isolated counters regardless of which `SyncFlows` instance or how many threads run concurrently.
- A single aggregate WARN is emitted at the end of the run when `unresolved > 0`: quieter wording for a partial failure, escalated wording (naming the likely causes) when every lookup in the run failed.
- Scope is `io.kestra.plugin.git.SyncFlows` only (OSS); no `plugin-git-lib` change, since `fetchFlowFromApi`/`FlowLookup` are private to this task and `plugin-ee-git` has no flow-sync counterpart.

closes: https://github.com/kestra-io/plugin-git/issues/336

## Test plan

- [x] `rtk test ./gradlew test` passes (`SyncFlowsTest`: 33/33; full suite: 96/97, the 1 pre-existing failure is the unrelated `CloneTest.cloneSelfHostedGiteaRepo` case requiring a local Gitea container not started in this session)
- [x] Added tests: dry-run single forced-500 lookup asserts `unresolvedFlowLookups == 1` (extended existing test); dry-run with every fixture flow forced to 500 asserts `unresolvedFlowLookups == 3` and the escalated aggregate WARN; clean dry-run against new (404) and existing (200) flows asserts `unresolvedFlowLookups == 0`; non-dry-run post-import re-fetch failure asserts the counter is incremented too; concurrent invocation of the same task instance from two threads asserts each invocation's own output reflects only its own lookups (verified this test fails against the previous shared-instance-field implementation and passes against the `ThreadLocal`-based fix)

---
*[View as Artifact](https://claude.ai/artifact/P58jAfoNF2Tr2j6cwFfUZQ)*


jymaire:
## QA report — SyncFlows `unresolvedFlowLookups` (#336)

**Edition**: EE | **Docker tag**: `develop` (next: 2.0.1)
**Instance**: http://z9weq.kestra.docker.localhost:1355/ (left running for manual review — `docker rm -f kestra-z9weq` to stop)
**Screenshots**: https://claude.ai/artifact/6EjkBhMSoUVEkeuYCDnVA3

### Summary

| Scenario | Result | `unresolvedFlowLookups` |
|----------|--------|------------------------|
| `git_dryrun` — SyncFlows dry-run, clean run | ✅ SUCCESS | **0** |
| `git_apply` — SyncFlows real import (non-regression) | ✅ SUCCESS | **0** |
| Systemic non-404 failure → non-zero count + escalated WARN | ⚙️ unit-tested | n/a in browser |

The new `unresolvedFlowLookups` output key is present in the loaded plugin schema and surfaces on the task output at runtime, reporting **0** on healthy runs (verified via the execution **Input/Output → Debug Expression** panel: `{{ outputs.sync.unresolvedFlowLookups }}` → `0`).

The **non-zero + aggregate-WARN** path requires a systemic non-404 failure of the single-flow Kestra API lookup, which cannot be induced through the browser against a healthy instance (every lookup resolves 200/404, so the count is always 0). That path is covered by the PR's unit tests (`SyncFlowsTest`, 32/32), including a run with every fixture flow forced to HTTP 500 asserting the count and the escalated WARN.

> Both flows sync from a local bare git repo (`file:///tmp/kestra-qa-336/repo.git`, one `flows/hello_qa.yaml`) to keep QA self-contained. The example repos in the task's `@Example` blocks (`github.com/kestra-io/flows`, `anna-geller/product`) no longer exist — worth refreshing in a follow-up, unrelated to this PR.

### Flow 1: `git_dryrun` (✅ SUCCESS)

<details>
<summary>Flow YAML</summary>

```yaml
id: git_dryrun
namespace: qa.git
tasks:
  - id: sync
    gitDirectory: flows
    targetNamespace: qa.gitsync
    includeChildNamespaces: true
    url: file:///tmp/kestra-qa-336/repo.git
    branch: main
    dryRun: true
    kestraUrl: http://localhost:8081
    auth:
      username: admin@kestra.io
      password: Kestra42!   # local dev default on an ephemeral instance
```
</details>

**Gantt** ([screenshot](https://claude.ai/artifact/6EjkBhMSoUVEkeuYCDnVA3#gitsync_dry-gantt))
| Task | Status | 
|------|--------|
| `sync` | SUCCESS |
| **Total** | **SUCCESS** |

**Outputs synthesis** ([screenshot](https://claude.ai/artifact/6EjkBhMSoUVEkeuYCDnVA3#gitsync_dry-outputs))
`outputs.sync.unresolvedFlowLookups = 0`; `outputs.sync.flows = kestra:///…/gitsync-dry/…` (diff ION file present). No aggregate WARN emitted (correct — nothing unresolved).

### Flow 2: `git_apply` (✅ SUCCESS — non-regression)

<details>
<summary>Flow YAML</summary>

```yaml
id: git_apply
namespace: qa.git
tasks:
  - id: sync
    gitDirectory: flows
    targetNamespace: qa.gitsync
    includeChildNamespaces: true
    url: file:///tmp/kestra-qa-336/repo.git
    branch: main
    dryRun: false
    kestraUrl: http://localhost:8081
    auth:
      username: admin@kestra.io
      password: Kestra42!
```
</details>

**Outputs synthesis** ([screenshot](https://claude.ai/artifact/6EjkBhMSoUVEkeuYCDnVA3#gitsync_apply-outputs))
`outputs.sync.unresolvedFlowLookups = 0`; `outputs.sync.flows` present. The flow `hello_qa` was really imported into `qa.gitsync` (GET `/api/v1/main/flows/qa.gitsync/hello_qa` → 200) — confirms the write path still works and the counter stays 0 on a real sync.

### Notes
- No timeouts; both executions completed in <1s.
- Environment aside (not a PR issue): this EE `develop` build deprecates `kestra.security.super-admin` in favor of `kestra.security.instance-owner`, and enforces tenant id `main` for API operations — the QA instance was reconfigured accordingly.

_Verdict: **PASS** — the new output signal works as specified on healthy runs; the failure-path behavior is unit-test-covered._

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -171,11 +172,53 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
...
+    private transient AtomicInteger unresolvedFlowLookups;
+
     @Override
     public Property<String> fetchedNamespace() {
         return this.targetNamespace;
     }
 
+    @Override
+    public Output run(RunContext runContext) throws Exception {
+        this.attemptedFlowLookups = new AtomicInteger(0);
```
jymaire: 🟠 [Performance/Thread-safety] Reset-in-run() protects sequential reuse only, not concurrent reuse
Problem: attemptedFlowLookups/unresolvedFlowLookups are shared instance fields. Resetting them here at the top of run() guards against one execution's stale counts leaking into the *next* sequential run() on the same SyncFlows object, but if the same object instance is invoked concurrently (e.g. as the templated child task of a parallel loop over several namespaces — see this class's own second @Example, which loops SyncFlows over 3 namespaces sequentially today but the same shape applies if that loop is ever made parallel — or two overlapping executions of the same flow reusing a cached Task object), one thread's `new AtomicInteger(0)` reassignment here can clobber the reference another concurrently-running thread's fetchFlowFromApi()/output() is actively reading/incrementing (lines 441, 460, 496). The counts and the escalated-vs-partial WARN wording can then be misattributed across the two concurrent runs, silently producing a wrong unresolvedFlowLookups value — undermining the exact trustworthy-signal goal of #336. The in-code comment above (and the PR description) both claim this handles "a task instance reused across concurrent executions"; that guarantee does not actually hold.
Fix: move this state out of shared instance fields — e.g. a ThreadLocal<AtomicInteger[]> set here and removed in a finally block after output() has read it, so each concurrent call on the same object instance owns its own counters.

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -171,11 +172,53 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
...
+        this.unresolvedFlowLookups = new AtomicInteger(0);
+
+        Output output = super.run(runContext);
+
+        int attempted = this.attemptedFlowLookups.get();
```
jymaire: 🟢 [Guidelines] Use var for local variables
Problem: `Output output`, `int attempted`, `int unresolved` use explicit types where the guideline mandates `var` for all local variables whenever legally possible.
Fix: `var output = super.run(runContext);`, `var attempted = this.attemptedFlowLookups.get();`, `var unresolved = this.unresolvedFlowLookups.get();`

jymaire (commented):
## Kestra Plugin Code Review

Issue #336 asked for: (1) a count/flag on `SyncFlows.Output` surfacing unresolved single-flow lookups, (2) a docs pass on the new output, (3) tests, and (4) consideration of a louder signal when a run is fully unresolved. All four are present: `unresolvedFlowLookups` (Integer, documented with `@Schema`), a class-level `@Schema` pointer to it, an escalated vs. partial aggregate WARN, and dedicated tests for the 404-exclusion, partial-failure, fully-unresolved, and non-dry-run re-fetch cases.

- 🟢 Nit — `Output output = super.run(runContext);`, `int attempted = ...`, `int unresolved = ...` in the overridden `run()` should be `var` per "Java Style & Local Variables": *use var for all local variables whenever legally possible*.

### Security (OWASP Top 10:2025 + KPS) — 0 findings
✅ No issues found. No new external input, no secrets, no crypto, no new dependency, no injection surface. The new field is a plain count with no sensitive content.

### Performance — 1 finding
- 🟠 Thread-safety — the two `transient AtomicInteger` counters are shared *instance* fields reassigned at the top of the overridden `run()` and read again after `super.run()` / inside `fetchFlowFromApi`/`output()`. That protects against **sequential** reuse of the same `SyncFlows` instance (one execution finishes, then another starts), but not **concurrent** reuse of the same instance — e.g. the same templated child-task object invoked once per iteration by a parallel loop over several namespaces (the class's own second `@Example` loops `SyncFlows` over 3 namespaces — swap that for a parallel construct and you have exactly this), or two overlapping executions of the same flow. Under real concurrency, one thread's `this.attemptedFlowLookups = new AtomicInteger(0)` can clobber the reference another concurrently-running thread is actively incrementing, and the final `.get()` reads whichever object happens to be referenced at that moment — counts and the escalated/partial WARN wording can end up misattributed between the two concurrent runs. This directly undercuts the trustworthiness of the very signal #336 asked for. The PR description and the in-code comment both assert this reset pattern "never leaks one execution's counts into another's" under concurrent reuse — that claim is only true for sequential reuse.
  Fix: don't keep this as instance state. Use a `ThreadLocal<AtomicInteger[]>` (or similar) set at the top of `run()` and removed in a `finally` block once `output()` has read it — matching the existing "ThreadLocal values not removed after task completion" guidance — so each concurrent invocation truly owns its own counters regardless of what any other concurrent call on the same object is doing.

**Verdict:** REQUEST CHANGES

src/test/java/io/kestra/plugin/git/SyncFlowsTest.java
```diff
@@ -833,6 +841,197 @@ void dryRun_whenFlowApiFailsWithNon404Status_shouldReportNoRevisionInsteadOfFabr
...
+     * server's forced-failure state is keyed only by namespace/flowId (identical for both concurrent invocations
+     * of this task instance, since they target the same namespace), and there is no hook to force interleaving at
+     * the exact instant the old code reassigned the shared counter field. This test instead runs the SAME task
+     * instance concurrently from two threads and asserts each invocation's own output reflects only its own
+     * lookups (3 attempted, 1 unresolved). Per-invocation isolation (this task uses a ThreadLocal) makes that
+     * assertion hold deterministically, whereas the old shared-field approach could let one thread's reset or
+     * increments bleed into the other's final count under real thread interleaving.
+     */
+    @Test
+    void run_concurrentInvocationsOnSameInstance_shouldKeepIndependentUnresolvedCounts() throws Exception {
```
jymaire, line 995: 🟡 [Test quality] Concurrency test is a directional guard, not a proof
Problem: The test comment correctly notes there is no deterministic interleaving hook, so a green run does not guarantee the old shared-field bug would always have been caught (both threads would need to race into the same narrow window for the old code to fail this specific assertion).
Fix: No action required — the ThreadLocal fix removes the race by construction. Just keeping this as a note so a future reader does not over-trust this test as a strict regression guard for the exact race; it is best-effort.

jymaire (commented):
## Kestra Plugin Code Review (round 2)

Re-review of commit 9b70696, addressing the round-1 thread-safety finding (transient `AtomicInteger` instance fields) and the `var` nit.

Issue #336: `unresolvedFlowLookups` surfaces on `Output`, 404 is still excluded from "unresolved", and the aggregate WARN escalates wording when every lookup fails vs. a partial failure. Unchanged from round 1, still satisfied.

### Kestra Guidelines — 0 findings
- `var` nit from round 1 is fixed: `run()` now uses `var output`, `var counters`, `var attempted`, `var unresolved`.
- `LOOKUP_COUNTERS` is `static`, so Lombok's `@ToString`/`@EqualsAndHashCode` on `SyncFlows` already exclude it by default (static fields aren't included) — no explicit `@ToString.Exclude` needed here.
- The Javadoc block on `LOOKUP_COUNTERS` documents a genuinely non-obvious design choice (ThreadLocal vs. instance field, and why), which fits the "hidden constraint / subtle invariant" exception in the comments guideline.
- ✅ No other violations found.

### Security (OWASP Top 10:2025 + KPS) — 0 findings
✅ No issues found. No secrets, no new external calls, no injection surface introduced by this batch.

### Performance / Thread-safety — round-1 finding resolved
Verified against `AbstractSyncTask.run()` in plugin-git-lib (build.gradle pins `1.0.2`, matching the local checkout): resource processing uses a plain sequential `.stream()` (not `parallelStream()`), and there is no `ExecutorService`/`CompletableFuture`/`new Thread` anywhere in the shared kernel (`AbstractGitTask`, `AbstractCloningTask`, `AbstractSyncTask`, `GitService`, `CloneService`). So `simulateResourceWrite` / `writeResource` (which call `fetchFlowFromApi`) and the final `output(...)` call all run synchronously on the same thread that invoked `SyncFlows.run()`.

Given that, the `ThreadLocal<LookupCounters>` fix is correct:
- `LOOKUP_COUNTERS.set(new LookupCounters())` at the top of `run()` and every subsequent `.get()` (in `fetchFlowFromApi` and `output(URI)`) execute on the same thread, so there is no path where `.get()` observes an empty ThreadLocal / NPEs. `fetchFlowFromApi` is private and only ever reached from inside the `try` block of `SyncFlows.run()`, confirmed via grep — no test or other call site bypasses that window.
- `LOOKUP_COUNTERS.remove()` is in a `finally`, so it also runs on exception paths (e.g. a thrown `KestraRuntimeException`/`ApiException` from deeper in `super.run()`), no leak on a pooled worker thread.
- Because the field is a `static ThreadLocal` rather than an instance field, concurrent invocations — same instance or different instances — never share state; correctness no longer depends on sequencing within a single instance, which is a strictly stronger guarantee than the round-1 fix needed.
- Correctness preserved: 404 still excluded from `unresolved` (only `attempted` increments unconditionally), both the dry-run (`simulateResourceWrite`) and non-dry-run (`writeResource`'s post-import re-fetch) paths increment through the same `fetchFlowFromApi`, `unresolvedFlowLookups` is always populated (never null) and 0 on a clean run per the new tests, and the WARN wording still distinguishes "all lookups failed" (degraded/permissions wording) from "partial failure" (quieter wording).

### Test quality
The new tests correctly cover: 404 exclusion (`shouldReportZeroUnresolvedLookups` with new-to-Kestra flows), a clean run (0 unresolved), and the non-dry-run re-fetch path (`write_whenPostImportRefetchFailsWithNon404Status_shouldCountUnresolvedLookup`) — closing the "both paths counted" gap.

🟡 Suggestion (non-blocking) on `run_concurrentInvocationsOnSameInstance_shouldKeepIndependentUnresolvedCounts`: the test comment itself acknowledges it can't force a deterministic interleaving, which is honest, but worth being explicit that this makes the test a directional regression guard rather than a proof — with the old shared-field bug, both concurrent invocations would need to race into the *same* window for the assertion to actually fail, so a green run today doesn't guarantee the old bug would always be caught either. No action required since the ThreadLocal design removes the race by construction rather than relying on the test to catch it.

**Intended verdict:** APPROVE — posted as COMMENT because GitHub blocks self-review APPROVE/REQUEST_CHANGES for the PR author.

src/main/java/io/kestra/plugin/git/SyncFlows.java
jymaire: Addressed in 9b70696: the counters are now a `private static final ThreadLocal<LookupCounters>` set at the top of `run()` and cleared in a `finally`, giving true per-invocation isolation (correct under concurrent same-instance reuse). The in-code comment and PR description were corrected to drop the inaccurate 'concurrent executions' claim.

src/main/java/io/kestra/plugin/git/SyncFlows.java
jymaire: Applied in 9b70696 — `run()` locals now use `var`.

src/test/java/io/kestra/plugin/git/SyncFlowsTest.java
jymaire: Acknowledged — the concurrency test is an intentional directional guard (documented in its Javadoc). It was verified to fail 5/5 against the pre-fix field-reassignment implementation and pass against the ThreadLocal fix; a fully deterministic race isn't feasible in this harness (the mock server's forced-failure state isn't per-invocation and there's no hook to force interleaving at the reset instant). Keeping as-is.

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -171,11 +171,59 @@ public class SyncFlows extends AbstractSyncTask<Flow, SyncFlows.Output> {
...
 
+    /**
+     * Per-invocation counters for single-flow API lookups (fetchFlowFromApi). Held in a static
+     * ThreadLocal — rather than an instance field reset at the top of run() — so that concurrent
+     * reuse of the same SyncFlows instance (e.g. the same templated child task invoked once per
+     * parallel-loop iteration, or two overlapping executions) never lets one invocation's reset or
+     * increments bleed into another's count. Each invocation sets its own counters in run() and
+     * clears them in a finally block once output() has read them.
+     */
+    private static final ThreadLocal<LookupCounters> LOOKUP_COUNTERS = new ThreadLocal<>();
```
fdelbrayelle: 🟡 [Performance] ThreadLocal NPE risk if AbstractSyncTask ever parallelizes
Problem: `LOOKUP_COUNTERS` relies on the implicit assumption that `fetchFlowFromApi()` always runs on the same thread that called `SyncFlows.run()`. That's true today because plugin-git-lib's `AbstractSyncTask.run()` processes resources sequentially, but it's an internal detail of a separately-versioned shared dependency this repo doesn't control — a future parallel stream there would NPE on `LOOKUP_COUNTERS.get()`.
Fix: `ThreadLocal.withInitial(LookupCounters::new)` instead of `new ThreadLocal<>()` — self-initializes on any thread at near-zero cost, removing the coupling to plugin-git-lib's current single-threaded implementation.

fdelbrayelle (approved):
## Kestra Plugin Code Review

Issue #336 asked for: (1) surface unresolved single-flow lookups on `SyncFlows.Output` as a count/flag, (2) consider a louder warning when every lookup fails, (3) docs + tests for the new output key. All three are delivered: `unresolvedFlowLookups` (Integer, always populated) on `SyncFlows.Output`, an escalated aggregate WARN when `unresolved == attempted` vs. a quieter partial-failure WARN, a `@Schema` doc entry with `"""` block per convention, and 5 new tests covering the happy path, partial failure, full failure, the non-dry-run re-fetch path, and concurrent reuse of the same task instance.

Correctly scoped to `io.kestra.plugin.git.SyncFlows` only, with no change to `plugin-git-lib`'s `AbstractSyncTask` — the PR description explicitly justifies this (`fetchFlowFromApi`/`FlowLookup` are private to this task; `plugin-ee-git` has no flow-sync counterpart), which is the right call per AGENTS.md's "shared plumbing lives in plugin-git-lib, only concrete tasks belong here."

### Kestra Guidelines — 0 blocking findings
✅ `@Schema` present on the new `Output.unresolvedFlowLookups` field, using a `"""` text block per the documentation formatting rules, and the class-level `@Schema(description = ...)` was updated to point at it.
✅ `var` used consistently for new local variables (`output`, `counters`, `attempted`, `unresolved`).
✅ No new task input property was added, so group/`@PluginProperty`/`@Min`/`@Max` rules don't apply here — only an output field.
🟢 Nit: the Javadoc-style comment on `LOOKUP_COUNTERS` (7 lines) and the per-test rationale comments are longer than the "one short line max" default, but each documents a genuinely non-obvious concurrency invariant (why `ThreadLocal` over an instance field, why the race can't be deterministically reproduced in the test harness), which the guidelines' "non-obvious algorithm" exception covers. Not requesting a change, just flagging it's close to the line — consider trimming to the single sentence that matters if this pattern is reused elsewhere.

✅ No new dependency, no HTTP client behavior change, no credential handling touched — the diff only adds counters and log/output surface around an existing, already-reviewed `fetchFlowFromApi` failure path.
✅ Log statements only include namespace/flow id/status code (pre-existing pattern), no secret-named variables introduced.
✅ No new deserialization, crypto, or process-execution surface.

### Performance — 1 suggestion
🟡 **Suggestion** — `SyncFlows.java:182` / `SyncFlows.java:446`: fragile coupling to `AbstractSyncTask.run()`'s current sequential (non-parallel) stream processing.
Problem: `LOOKUP_COUNTERS` is a `static ThreadLocal` populated only by the overridden `run()` on the calling thread; `fetchFlowFromApi()` assumes it always runs on that same thread and calls `LOOKUP_COUNTERS.get().attempted++` without a null check. This holds today because `plugin-git-lib`'s `AbstractSyncTask.run()` processes the resource stream sequentially and calls `output()` synchronously on the same thread — but that's an internal implementation detail of a separately-versioned shared library this repo doesn't control (AGENTS.md: bump the lib version to pick up plugin-git-lib changes). If a future `plugin-git-lib` version parallelizes that stream (e.g. `.parallel()` for large namespaces) for performance, `fetchFlowFromApi` would NPE on `LOOKUP_COUNTERS.get()` on any worker thread that didn't go through `SyncFlows.run()`.
Fix: use `ThreadLocal.withInitial(LookupCounters::new)` instead of `new ThreadLocal<>()`. This costs nothing today and makes the counters self-initializing regardless of which thread reaches `fetchFlowFromApi`, removing the dependency on an unenforced single-thread assumption in an external, independently-versioned dependency.

**Verdict:** APPROVE

src/main/java/io/kestra/plugin/git/SyncFlows.java
jymaire: Applied in 8adcce6: `LOOKUP_COUNTERS` is now a `private static final ThreadLocal<LookupCounters>` created via `ThreadLocal.withInitial(LookupCounters::new)`, so `fetchFlowFromApi()` gets a fresh counter on any thread that didn't go through `run()` — no NPE even if a future `plugin-git-lib` version parallelizes the resource stream. Costs nothing today and removes the unenforced single-thread assumption. `SyncFlowsTest` green (18/18) against the change. Thanks for the catch!

Issue #16, opened 2023-11-27
New `git.Push` task allowing to push namespaceFiles and custom inputFiles to a Git branch


anna-geller:
## Motivation 

Building a UI-only feature to push Namespace Files to Git is against Kestra's product mindset of `Everything as code and from the UI`. We want to support the use case of pushing to Git from Kestra. However, using a dedicated task instead of a UI-only feature.

With the help of inputs for defining a Git repository and credential details + creating Blueprints for various Git systems, we can provide a fairly intuitive and simple solution to the problem of "Push to Git from Kestra" without compromising the philosophy of Everything as Code.

## Task spec

Example flow with extra Clone step:
```yaml
id: flow
namespace: dev

inputs:
  - name: commit_message
    type: STRING
    defaults: my commit message

tasks:
  - id: wdir
    type: io.kestra.core.tasks.flows.WorkingDirectory
    
    tasks:
      - id: clone
        type: io.kestra.plugin.git.Clone
        branch: feature_branch
        url: https://github.com/kestra-io/scripts
        username: anna-geller
        password: myPAT

      - id: push
        type: io.kestra.plugin.git.Push
        namespaceFiles:
          enabled: true
        inputFiles: null # optional list of extra files
        description: Adds all namespace files, commits and pushes them
        commitMessage: "{{ inputs.commit_message }}"
```

An example of a simpler flow that does both: clones the repo, adds files, commits and pushes them, all in one go.

```yaml
id: flow
namespace: dev

inputs:
  - name: commit_message
    type: STRING
    defaults: my commit message

  - name: branch
    type: STRING
    defaults: main

tasks:
  - id: git
    type: io.kestra.plugin.git.Push
    namespaceFiles:
      enabled: true
    inputFiles: null # optional map property with extra files --> myfile.py: print("hello")
    branch: feature_branch # mandatory, we'll create the branch if not exists
    url: https://github.com/kestra-io/scripts
    username: anna-geller
    password: mypat
    commitMessage: "{{ inputs.commit_message }}" # required
```

**Benefits**:

- Everything as code
- easy to manage + extend across all Git systems (GitHub, Bitbucket, Gitlab, self-hosted Git etc.)
- Users can push to a feature branch regularly and implement their own Git approval/merge strategy

**Under the hood:**

- we clone a repository
- we add `inputFiles` and `namespaceFiles` if specified on the task
- we commit and push to Git

PR #354, opened 2026-09-16
fix(git): stop fabricating revision 1 on unresolved SyncFlow lookup

### LINKED ISSUE #335: SyncFlow treats any API failure as "flow does not exist", fabricating revision 1
Follow-up from the review of #333 (fix for #327).

### Problem

`SyncFlow.java:161-166` catches every `ApiException` from `flows().flow(...)` and treats it as "the flow does not exist":

- a genuine `404` (flow not yet synced) — correct
- a `401`/`403` (bad, expired, or under-scoped token) — **wrong**
- a `5xx` from the Kestra API — **wrong**

All three collapse into the same branch, which then fabricates `projectedRevision = 1`. On the dry-run path this means a transient auth or API outage silently reports an existing flow (say, at revision 7) as newly `ADDED` at revision 1, with no trace of why. The user sees a plausible-looking diff that is simply wrong.

This is the same defect class as #327 itself: a single `null`/exception branch standing in for two distinct states.

### Fix applied in `SyncFlows` (for reference)

#333 fixed the equivalent code in `SyncFlows.fetchFlowFromApi` by:

1. narrowing the "absent" branch to `e.getCode() == 404` only;
2. logging a `WARN` with namespace, flow id, and status for any other code;
3. returning a `FlowLookup(flow, resolved)` record so callers can distinguish "flow is absent" from "lookup failed" and avoid fabricating a revision in the latter case.

### Suggested scope

Apply the same narrowing to `SyncFlow`. Since both tasks now need "resolve a single flow, distinguishing 404-absent from a real API failure" — and both also carry a duplicated `toNamedTempFile` helper — consider extracting the shared logic rather than copying the fix a second time.

Raised by @fdelbrayelle in review of #333; kept out of that PR to keep it scoped to #327.

jymaire:
## Summary

- `SyncFlow`'s dry-run path treated any `ApiException` other than 404 as "flow does not exist", silently falling through to `projectedRevision = 1` even for real API failures (401/403/500, or code 0 meaning no HTTP response at all — e.g. connection/DNS/TLS failure). Dry-run now **fails fast** instead, throwing a `KestraRuntimeException` that names the namespace, flow id, HTTP status (or "no HTTP response") and points at the two realistic causes (the `auth` token can't read flows in this namespace, or `kestraUrl` points at the wrong instance).
- Extracted the 404-vs-real-failure single-flow lookup into a new package-private helper `io.kestra.plugin.git.FlowApiLookup` (alongside `SyncFlow`/`SyncFlows`, same shape as other non-task helpers in this package) so the fix lives in one place. `SyncFlows.fetchFlowFromApi` and its private `FlowLookup` record were removed and replaced with calls to the shared helper — `SyncFlows`' own behavior (unresolved lookups keep reporting `UNCHANGED`/no revision rather than throwing, per its own contract) is unchanged, verified by both pre-existing tests and a new regression test.
- 404 during dry-run still projects revision 1 (new flow); an existing flow still projects `revision + 1`. The non-dry-run path (post-import re-fetch) is untouched.

## ⚠️ Merge-conflict note

Branch `feat/349-namespace-directory` (in flight, implementing #336) adds `attempted`/`unresolved` `AtomicInteger` counters inside `SyncFlows.fetchFlowFromApi` (incremented on entry and in the non-404 branch). Since this PR extracts that method into `FlowApiLookup`, **whoever merges second must reconcile the counter increments with the extraction** (most likely by moving the counter increments to the `SyncFlows` call sites around `FlowApiLookup.fetch(...)`, since the helper itself is intentionally state-free).

closes: https://github.com/kestra-io/plugin-git/issues/335

## Test plan

- [x] `rtk test ./gradlew test` passes (full suite, including the two new `SyncFlowTest` failure-status regression tests for 401/403/500, the existing 404/200 dry-run cases, and a new `SyncFlowsTest` regression guard for the extraction)

---
*[View as Artifact](https://claude.ai/artifact/5mAUaGYJijGsbd4bu1Dtam)*


jymaire:
## QA Report — browser-based (Kestra EE)

**Edition**: EE | **Docker tag**: `develop` (next release; EE dev `2.0.1`, plugin builds against `kestraVersion 2.0.2`)
**Instance**: http://qpm6m.kestra.docker.localhost:1355/ui/ — left running for manual review (`docker rm -f kestra-qpm6m` to stop)
**Screenshots**: https://claude.ai/artifact/WBbWQK9ovJkuh7uhQohh7i
**Plugin build**: `plugin-git-4.0.2-SNAPSHOT.jar` built from `fix/335-syncflow-lookup-failure`, deployed to the mounted plugins dir.
**Git source**: local Gitea (`gitea-test`, self-signed HTTPS on `:3443`) with a seeded repo `gitea_admin/kestra-test` containing `flows/hello.yml`; the container trusts the Gitea CA via `trustedCaPemPath`.

### Result summary

| # | Scenario | Path exercised | Result |
|---|----------|----------------|--------|
| 1 | Dry-run, **new** flow, valid `kestraUrl`+auth | 404-absent → project revision 1 | ✅ SUCCESS, `revision = 1` |
| 2a | Import (`dryRun:false`) of a new flow | real import | ✅ SUCCESS, `revision = 1` |
| 2b | Dry-run against the now-**existing** flow | existing → project revision+1 | ✅ SUCCESS, `revision = 2` |
| 3 | Dry-run, **unreachable** `kestraUrl` | API failure → fail fast (no fabricated revision) | ✅ FAILED (task errors, no bogus `revision:1`) |

**Verdict: PASS.** The dry-run revision projection is correct on the happy paths (new→1, existing→2), and an unreachable Kestra API no longer yields a plausible-but-wrong success — the execution fails. See the scope note under Scenario 3 for exactly which code path the browser reached vs. what the unit tests cover.

### Scenario 1 — dry-run, new flow (non-regression, 404 path)

```yaml
id: qa_sf_dryrun_new
namespace: qa.test
tasks:
  - id: sync
    type: io.kestra.plugin.git.SyncFlow
    url: https://localhost:3443/gitea_admin/kestra-test.git
    branch: main
    flowPath: flows/hello.yml
    targetNamespace: qa.dryrunnew
    dryRun: true
    trustedCaPemPath: /tmp/kestra-git-qa/cert.pem
    kestraUrl: http://localhost:8082
    auth:
      auto: false
      username: admin@kestra.io
      password: "<redacted — local dev default>"
```
</details>

**Gantt** ([shot](https://claude.ai/artifact/WBbWQK9ovJkuh7uhQohh7i#qa_sf_dryrun_new-gantt)) — `sync` SUCCESS.
**Outputs** ([shot](https://claude.ai/artifact/WBbWQK9ovJkuh7uhQohh7i#qa_sf_dryrun_new-outputs)) — `{{ outputs.sync.revision }}` → **1**. The 404 "flow does not exist" branch still correctly projects revision 1; the fix did not regress it.

### Scenario 2 — import then dry-run existing (existing-flow path)

<details><summary>Flow YAMLs</summary>

```yaml
# 2a — import
id: qa_sf_import
namespace: qa.test
tasks:
  - id: sync
    type: io.kestra.plugin.git.SyncFlow
    url: https://localhost:3443/gitea_admin/kestra-test.git
    branch: main
    flowPath: flows/hello.yml
    targetNamespace: qa.imported
    dryRun: false
    trustedCaPemPath: /tmp/kestra-git-qa/cert.pem
    kestraUrl: http://localhost:8082
    auth: { auto: false, username: admin@kestra.io, password: "<redacted>" }
---
# 2b — dry-run against the now-existing flow (same targetNamespace)
id: qa_sf_dryrun_existing
namespace: qa.test
tasks:
  - id: sync
    type: io.kestra.plugin.git.SyncFlow
    url: https://localhost:3443/gitea_admin/kestra-test.git
    branch: main
    flowPath: flows/hello.yml
    targetNamespace: qa.imported
    dryRun: true
    trustedCaPemPath: /tmp/kestra-git-qa/cert.pem
    kestraUrl: http://localhost:8082
    auth: { auto: false, username: admin@kestra.io, password: "<redacted>" }
```
</details>

**2a import** ([shot](https://claude.ai/artifact/WBbWQK9ovJkuh7uhQohh7i#qa_sf_import-gantt)) — SUCCESS, `outputs.sync.revision = 1`; `qa.imported/hello_qa` created at revision 1 (verified via API).
**2b dry-run** ([Gantt](https://claude.ai/artifact/WBbWQK9ovJkuh7uhQohh7i#qa_sf_dryrun_existing-gantt) · [Outputs](https://claude.ai/artifact/WBbWQK9ovJkuh7uhQohh7i#qa_sf_dryrun_existing-outputs)) — SUCCESS, `{{ outputs.sync.revision }}` → **2**. Existing flow at rev 1 correctly projects the next revision (2).

### Scenario 3 — dry-run with an unreachable `kestraUrl` (fail-fast)

```yaml
id: qa_sf_badurl
namespace: qa.test
tasks:
  - id: sync
    type: io.kestra.plugin.git.SyncFlow
    url: https://localhost:3443/gitea_admin/kestra-test.git
    branch: main
    flowPath: flows/hello.yml
    targetNamespace: qa.imported
    dryRun: true
    trustedCaPemPath: /tmp/kestra-git-qa/cert.pem
    kestraUrl: http://localhost:1     # unreachable → no HTTP response (ApiException code 0)
    auth: { auto: false, username: admin@kestra.io, password: "<redacted>" }
```
</details>

**Gantt** ([shot](https://claude.ai/artifact/WBbWQK9ovJkuh7uhQohh7i#qa_sf_badurl-gantt)) — `sync` **FAILED**.
**Logs** ([shot](https://claude.ai/artifact/WBbWQK9ovJkuh7uhQohh7i#qa_sf_badurl-logs)) — `Connect to http://localhost:1 failed: Connection refused` → `ApiException{code=0}`. The task errors out; it does **not** emit a green `revision: 1`.

> **Scope note (honest):** with a fully unreachable `kestraUrl`, the failure surfaces at the **first** dry-run API call — `validateFlows` (`SyncFlow.java:168`) — *before* execution reaches the single-flow lookup that #335 actually changed. So this scenario proves the end-to-end guarantee "a broken Kestra API makes dry-run fail rather than fabricate a revision", but it does not isolate the #335 branch itself (a bad URL would fail here with or without the fix). The exact fixed branch — `validateFlows` succeeds yet the single-flow lookup returns a **non-404** (401/403/500) → throw instead of fabricating `revision:1` — cannot be isolated through the browser without a token scoped to allow validate-but-deny-read, and is covered directly by the unit tests in this PR (`SyncFlowTest` 401/403/500 + the 404/200 regression guards, all green in CI).

### Environment notes (not defects in this PR)
- Hit the documented EE `develop` bootstrap gotcha: `kestra.security.super-admin` is deprecated in 2.0.x in favour of `kestra.security.instance-owner`, and tenant-scoped writes 403 until the instance-owner is bound to a tenant. Resolved by booting with `instance-owner` config on a fresh volume and creating tenant **`main`** via the Instance Owner UI (auto-binds Admin), then re-logging in. Flow **writes** must go through the UI editor (a cookie-only API session grants reads only).
- The `gitea-test` container name/port is fixed and host-wide (shared across worktrees); harmless here.

*QA performed on branch `fix/335-syncflow-lookup-failure` via `agent-browser` against the live EE instance above.*

src/main/java/io/kestra/plugin/git/SyncFlow.java
```diff
@@ -170,22 +170,23 @@ public Output run(RunContext runContext) throws Exception {
...
-                    );
-                }
-                projectedRevision = 1;
+            // Projected revision: existing + 1, or 1 for a new flow. A lookup that couldn't be resolved
+            // (anything other than a 200 or a 404) must not silently be treated as "flow does not exist" —
+            // that would fabricate revision 1 for what might be an existing flow, so dry-run aborts instead.
+            FlowApiLookup.FlowLookup lookup = FlowApiLookup.fetch(runContext, kestraClient, tenantId, rNamespace, flowId, false);
+            if (!lookup.resolved()) {
+                String statusDescription = lookup.statusCode() == 0 ? "no HTTP response" : "HTTP " + lookup.statusCode();
+                throw new KestraRuntimeException(
```
jymaire: 🟡 [Performance] git resource leak on unresolved-lookup throw
Problem: This throw exits run() before reaching git.close() at line 204, leaking the cloned repo's JGit file handles/working directory. Same gap already existed for the validation-failure throw above it, but this new throw site fires far more often now (any 401/403/500/no-response during dry-run used to fall through silently — now it throws).
Fix: Wrap the git clone lifecycle in try-with-resources (or a try/finally around the whole method body) so git.close() runs on every exit path, not just the successful one.

src/main/java/io/kestra/plugin/git/FlowApiLookup.java
```diff
@@ -0,0 +1,62 @@
...
+            }
+            // Only 404 means "flow does not exist"; any other status (including 0 for no HTTP response at
+            // all) is a real failure that leaves the lookup unresolved, so callers must not assume the flow
+            // is absent.
+            String abbreviatedMessage = StringUtils.abbreviate(StringUtils.defaultString(e.getMessage()), 500);
+            runContext.logger().warn(
+                "Failed to fetch flow {}.{} from the Kestra API (status {}): {}",
+                namespace, flowId, e.getCode(), abbreviatedMessage
+            );
+            return new FlowLookup(null, false, e.getCode(), abbreviatedMessage);
```
jymaire: 🟡 [Tests] No automated coverage for the code-0 "no HTTP response" branch
Problem: SyncFlowTest covers 401/403/500/404/200 via HttpServer, but statusCode == 0 (no HTTP response — connection/DNS/TLS failure), the exact case this PR's summary calls out by name, is only verified by manual EE QA (unreachable kestraUrl). A future refactor could silently regress the "no HTTP response" wording without any test catching it.
Fix: Add a unit test that forces an ApiException with code 0 (e.g., point kestraUrl at a closed local port so the HTTP client throws before getting a response) and assert the thrown KestraRuntimeException message contains "no HTTP response".

src/main/java/io/kestra/plugin/git/SyncFlows.java
```diff
@@ -394,33 +394,6 @@ private List<Flow> fetchFlowsFromKestra(KestraClient kestraClient, RunContext ru
         }
     }
```
jymaire, line 447: 🟢 [Guidelines] toNamedTempFile still duplicated
Problem: Issue #335 suggested extracting this alongside FlowApiLookup since both SyncFlow and SyncFlows carry an identical private toNamedTempFile helper; it's still duplicated verbatim in both files after this PR.
Fix: Non-blocking (issue phrased it as optional) — worth a quick follow-up extraction into a shared utility to avoid drift between the two copies.

jymaire (commented):
## Kestra Plugin Code Review

The PR does exactly what issue #335 asked: narrows SyncFlow's dry-run lookup to 404-only-means-absent, extracts the shared 404-vs-real-failure logic into `FlowApiLookup` (per the issue's suggested scope), and repoints `SyncFlows` at it with verified-unchanged behavior (existing 500 test + new 403 regression test). 404→revision 1 and existing-flow→revision+1 are preserved. The one piece of the issue's suggested scope left undone is the secondary, explicitly-optional ask to also dedupe `toNamedTempFile` — noted below as a non-blocking nit.

- 🟢 Nit — DRY: `toNamedTempFile` stays duplicated verbatim in both `SyncFlow.java` and `SyncFlows.java`. Issue #335 flagged this as a candidate for the same extraction as `FlowApiLookup` ("consider extracting the shared logic"). Not blocking since the issue phrased it as optional, but worth a fast follow so the two copies don't drift.
- 🟢 Nit — Unrelated `gradle.properties` kestraVersion bump (2.0.0 → 2.0.2) is bundled into this bug-fix PR (likely a rebase artifact). Doesn't touch `plugin-git-lib`/kestra-api-client (unchanged in this diff), so no SDK-signature note is needed, but scope-wise it's unrelated to issue #335.

- Error messages stay within the existing abbreviate(500) pattern before being logged/thrown (A10 / A09 compliant) — no raw exception text or secrets surfaced.
- `FlowApiLookup` is fully stateless (private ctor, static method, all state passed as parameters) — no shared mutable state, consistent with the concurrent branch's own note that "the helper itself is intentionally state-free."
- No new dependency, no template rendering of attacker-controlled input, no deserialization or crypto surface touched.

### Performance — 2 findings
- 🟡 Suggestion — Resource leak on the new throw path: in `SyncFlow.run()`, when the lookup is unresolved the method throws `KestraRuntimeException` before reaching `git.close()` at the end of the method, leaking the cloned repository's open file handles/working directory. This pattern already existed for the pre-existing validation-failure throw a few lines above, but this PR adds a second throw site with the same gap — and, because it now fires on any 401/403/500/no-response during dry-run (previously these cases fell through silently), it will be hit measurably more often. Consider wrapping the git lifecycle in try-with-resources or a try/finally so `git.close()` runs on every exit path.
- 🟡 Suggestion — Test gap on the code-0 branch: 401/403/500/404/200 are all covered by `HttpServer`-based unit tests, but the "no HTTP response" (status code 0) branch — the one added specifically to catch connection/DNS/TLS failures — has no automated regression test; it was only exercised by manual EE QA (unreachable `kestraUrl`). A unit test that forces an `ApiException` with code 0 (e.g., a `kestraUrl` pointing at a closed port) would lock in the "no HTTP response" wording and prevent a future refactor from silently reverting to a raw "HTTP 0" message.

**Verdict:** APPROVE

src/main/java/io/kestra/plugin/git/SyncFlow.java
jymaire: Fixed in dbbf06a — the clone is now wrapped in try-with-resources on `Git`, so the JGit repository is released on every exit path, including the dry-run unresolved-lookup throw. The old explicit `git.close()` was removed (no double-close).

src/main/java/io/kestra/plugin/git/FlowApiLookup.java
jymaire: Added in dbbf06a — `flowApiLookup_noHttpResponse_isUnresolvedWithStatusCodeZero` asserts `FlowApiLookup.fetch` maps a connection failure to `resolved=false`/`statusCode=0`. Tested at the FlowApiLookup level because in a dry run `validateFlows` and the single-flow lookup share the same `kestraUrl`, so an unreachable URL fails at `validateFlows` first.

src/main/java/io/kestra/plugin/git/SyncFlows.java
jymaire: Intentionally out of scope for #335. `toNamedTempFile` is duplicated across both editions (OSS + EE), so it belongs in `plugin-git-lib` (a lib bump), not this OSS-only fix — the issue lists it as an optional secondary suggestion. Tracking it separately to keep this PR scoped to the revision-fabrication defect.

Malaydewangan09:
The fix is needed. The class as written is not.

On placement: `plugin-git-lib` already owns `io.kestra.plugin.git.shared.services` (`GitService`, `CloneService`) and `AbstractKestraTask.kestraClient(...)`. A single-flow API lookup is the same category. Per AGENTS.md this repo holds concrete registered tasks and the plumbing lives in the lib, so `plugin-ee-git` picks up the same fix. As it stands this forks it into OSS only.

On shape: `FlowLookup(flow, resolved, statusCode, errorMessage)` carries two fields whose own javadoc says they are only meaningful when `resolved` is false, and `SyncFlows` ignores both. The helper logs a WARN and then `SyncFlow` immediately throws with the same namespace, flow id and status. `RunContext` is passed in only to reach a logger, into a class documented as state-free. `SyncFlow` passes `withSource=false` but still pays for a `core.FlowWithSource` build when all it reads is `getRevision()`.

The idiomatic version needs no record and no `runContext`:

```java
static Optional<FlowWithSource> fetch(KestraClient client, String tenantId, String ns, String id, boolean withSource) throws ApiException {
    try {
        return Optional.of(map(client.flows().flow(ns, id, tenantId, withSource, null, false)));
    } catch (ApiException e) {
        if (e.getCode() == 404) return Optional.empty();
        throw e;
    }
}
```

404 means absent, anything else propagates. `SyncFlow` wraps it in `KestraRuntimeException` at its call site with the message it wants, `SyncFlows` catches and warns to keep its tolerant contract. Each caller owns its own policy instead of negotiating it through a shared record.

Other points:

- The `try-with-resources` change is a real fix, `git.close()` was skipped on every exception path. But it reindents 65 lines and is the sole cause of the declared conflict with `feat/349-namespace-directory`. Worth splitting out.
- The asymmetry is codified rather than resolved. Same failure: `SyncFlow` aborts, `SyncFlows` reports `UNCHANGED` with no revision. Reporting `UNCHANGED` for a flow whose state is unknown is also a wrong diff, just a quieter one. If the premise of #335 is "no plausible-looking diff that is wrong", `SyncFlows` has the same bug.
- `dryRun_whenFlowApiFailsWith403_extractionShouldNotChangeBehaviour` duplicates the existing non-404 test with a different status code and tests the refactor rather than behaviour. The 401/403/500 trio should be one `@ParameterizedTest`, and they spin a raw `HttpServer` instead of the harness `SyncFlowsTest` already uses.
- Long dashes in the new comments and inside the user-facing exception message, against the repo writing rules. The exception text is four clauses of troubleshooting advice, which belongs in a log line rather than an exception. `StringUtils.defaultString(lookup.errorMessage())` is applied twice, the helper already defaulted it.
- Up to 500 chars of raw API error body land in an exception message that goes to execution logs. A 401 body from a proxy can echo request headers. Low risk, but durable.

Suggested split: land the narrowing in `SyncFlow` plus the `try-with-resources` fix as the bugfix, and move the shared lookup into `plugin-git-lib` separately so EE gets it too. If the extraction stays here for now, drop the record for `Optional` plus a propagated `ApiException`.

jymaire:
Thanks Malay, this is the right read on all counts. Adjusting the PR accordingly.

**Placement.** Agreed: the single-flow lookup is the same category as `GitService`/`CloneService` and `AbstractKestraTask.kestraClient(...)`, and per AGENTS.md it belongs in `plugin-git-lib` so `plugin-ee-git` inherits the fix instead of it forking OSS-only. I'll land it there as a shared service and consume it here via a version bump. That gates this PR on a lib `1.0.3` release, which is expected.

**Shape.** Dropping the record. The service becomes the `Optional<FlowWithSource> fetch(...)` you sketched: 404 -> empty, anything else propagates the `ApiException`. No `RunContext`, no status/error fields negotiated through a shared type. Each caller owns its policy: `SyncFlow` wraps it in `KestraRuntimeException` at the call site with the message it wants, `SyncFlows` catches and warns to keep its tolerant contract. `SyncFlow` also passes `withSource=false` so it stops paying for a source build it never reads.

**try-with-resources.** Agreed it is a real fix but orthogonal, and it is the sole cause of the `feat/349-namespace-directory` conflict. Splitting it into its own small bugfix PR and keeping this one focused on the revision fabrication.

**SyncFlows asymmetry.** Fair. Reporting `UNCHANGED` for a flow whose current state is unknown is a quiet wrong diff of the same shape as #335. `SyncFlows` should surface the unresolved lookup rather than fold it into a clean diff. That is exactly the `unresolvedFlowLookups` signal tracked in #336; I'll make `SyncFlows` here stop presenting unknown state as `UNCHANGED` and wire the signal in #336.

**Tests.** Collapsing the 401/403/500 cases into one `@ParameterizedTest` and moving them onto the `SyncFlowsTest` harness instead of a raw `HttpServer`.

**Writing rules / exception text / error body.** Removing the long dashes from the new comments and the user-facing message, shortening the exception to a terse throw with the troubleshooting advice moved to a log line, dropping the duplicated `StringUtils.defaultString`, and not spilling raw API error bodies (which can echo proxy headers) into an exception that reaches execution logs.

jymaire:
Pushed the reworked commit (`refactor(sync-flow): consume shared FlowLookupService, drop local lookup record`). Summary of what changed vs. the review:

- **Lookup moved to plugin-git-lib** as `shared.services.FlowLookupService` → kestra-io/plugin-git-lib#16. Consumed here via the `1.0.4` bump. `plugin-ee-git` inherits the same fix.
- **Record dropped.** The service returns `Optional<FlowWithSource>` (404 → empty) and propagates any other `ApiException`. `SyncFlow` fails fast with a terse `KestraRuntimeException` at the call site (advice moved to a WARN log, no raw API body), `SyncFlows` catches and warns to stay tolerant. `SyncFlow` now fetches with `withSource=false`. Dropped the duplicated `defaultString` and the long dashes.
- **try-with-resources split out** into #355, so this PR no longer carries the reindent that conflicted with `feat/349-namespace-directory`.
- **Tests consolidated.** `SyncFlow` 401/403/500 → one `@ParameterizedTest` on the shared `MockKestraApiServer` harness; `SyncFlows` non-404 trio collapsed into one (dropping the refactor-testing 403 duplicate); the `FlowApiLookup`-level test removed. `SyncFlowTest` + `SyncFlowsTest` green locally against `1.0.4-SNAPSHOT` (published to mavenLocal).
- **SyncFlows unresolved-state signal:** left tolerant here with a comment; surfacing it on the output as a first-class signal (so it can't read as a clean `UNCHANGED`) is #336's `unresolvedFlowLookups` work, to avoid overlapping that branch.

**Blocked on** the lib change: this consumes `plugin-git-lib:1.0.4`, so it needs kestra-io/plugin-git-lib#16 merged and `1.0.4` released before CI here can resolve the dependency (currently `1.0.4-SNAPSHOT` via mavenLocal). Marking as draft until then.

jymaire:
Rebased onto `main` and reworked now that #355 (try-with-resources) and #353 (unresolved-lookup signal) have merged. Force-pushed a single focused commit; the old diff is gone.

What this PR now does, on top of current `main`:
- **`SyncFlow` (single)** still fabricated `revision = 1` on any non-404 (the #335 bug is unfixed on `main`). It now routes the dry-run lookup through **`plugin-git-lib`'s shared `FlowLookupService` (1.1.0)** and fails fast with an actionable `KestraRuntimeException` instead of a wrong revision. Advice → WARN log; no raw API body in the exception.
- **`SyncFlows.fetchFlowFromApi`** now delegates its 404-vs-non-404 logic to the **same** `FlowLookupService`, removing the duplicated lookup while keeping the `LOOKUP_COUNTERS` and the `Output.unresolvedFlowLookups` signal from #353. Both editions share one lookup implementation, per the shared-kernel rule.
- Added a parameterized `SyncFlow` dry-run test (401/403/500) on the `MockKestraApiServer` harness.

`SyncFlowTest` (9) + `SyncFlowsTest` (18) green locally. Clean 4-file diff (`build.gradle`, `SyncFlow`, `SyncFlows`, `SyncFlowTest`).

**Only remaining blocker:** `build.gradle` consumes `plugin-git-lib:1.1.0`, which is released ([lib #16](https://github.com/kestra-io/plugin-git-lib/pull/16) is in `v1.1.0`) but not yet on Maven Central. I'll mark this ready and CI will pass once `1.1.0` is resolvable.

jymaire:
`plugin-git-lib:1.1.0` is now live on Maven Central, so the dependency resolves. Marked ready for review — CI should go green (SyncFlowTest + SyncFlowsTest pass locally against the released 1.1.0). @Malaydewangan09 over to you.

Malaydewangan09 (approved):

Malaydewangan09:
Follow-up after a second review pass. My earlier comment reviewed the branch state as it was then (`FlowApiLookup` in this repo). The merged form (`50af10d`) is different and addresses the substance: the helper moved to `plugin-git-lib` as `shared.services.FlowLookupService` with a `1.1.0` bump, it returns `Optional` and propagates `ApiException`, and the JGit `try-with-resources` was split into #355. The tests are now one `@ParameterizedTest` over `{401, 403, 500}` on the shared `MockKestraApiServer`. Good result, so the rest of this is small and against merged `main`.

Three things still open.

### The remediation text is in the log but not in the exception

`SyncFlow.java:185-193`. The WARN carries "check that 'auth' can read flows in this namespace and that 'kestraUrl' targets the intended instance". The `KestraRuntimeException` thrown two lines later does not:

```
"Cannot compute the projected revision for " + rNamespace + "." + flowId
    + ": the Kestra API single-flow lookup returned " + statusDescription
    + ". Dry-run aborted rather than reporting a possibly wrong revision."
```

A user whose dry-run fails on 403 sees the task error, which is the exception message, and gets no hint what to check. The PR description says the exception "points at the two realistic causes". It does not, only the log does. Build the string once and use it for both.

### The exception drops the cause

Same throw site. `new KestraRuntimeException(msg)` discards the `ApiException`. Every neighbouring throw passes it: `SyncFlow.java:230`, `SyncFlows.java:441`, `:444`, `:474`. Dropping `e.getMessage()` from the text is a deliberate and correct call, but dropping the cause too loses the stack for an unexplained 500 with no benefit. `new KestraRuntimeException(msg, e)`.

### The code 0 path has no test anywhere

The whole point of the fix is that a connection, DNS or TLS failure (`ApiException#getCode() == 0`, no HTTP response) must not fabricate revision 1. `SyncFlow.java:184` handles it, and nothing tests it. The parameterized test covers 401/403/500 only, and `FlowLookupService` ships in `plugin-git-lib` v1.1.0 with no test at all (the `services` package there has `CloneServiceTest` and `GitServiceTest`, no `FlowLookupServiceTest`). The earlier branch did have a direct code 0 test, it did not survive the move to the lib. Worth adding a `FlowLookupServiceTest` in the lib against an unreachable URL, which also gives that class its first test.

### Minor

- The PR body lists "a new `SyncFlowsTest` regression guard for the extraction". `50af10d` touches only `build.gradle`, `SyncFlow.java`, `SyncFlows.java` and `SyncFlowTest.java`. `SyncFlowsTest.java` has no diff. Stale text from the earlier branch, worth not carrying into the release notes.
- Issue #335 also asked to consider extracting the duplicated `toNamedTempFile` (`SyncFlow.java:220-232` and `SyncFlows.java:466-476`, still identical). It was phrased as "consider", so this is a note rather than a gap.
- `AGENTS.md` names every shared class in the "Shared kernel" section but was not updated with `FlowLookupService`. Cheap to keep current so nobody re-duplicates the lookup locally.

PR #357, opened 2026-09-21
feat(sync): allow separate source of truth for flows and Namespace Files

### LINKED ISSUE #237: Enable TenantSync and NamespaceSync to have separate source of truth for files
### Feature description

At the moment, TenantSync and NamespaceSync will sync all objects inside of Kestra based on the same source of truth. This works well for Kestra specific objects like Flows, Tests, Dashboards and Apps which are typically built directly inside of Kestra.

However, Namespace Files aren't typically modified inside of Kestra. Most use cases involve people wanting to take their already modified files (e.g. code) and sync them into Kestra so the workflow can use them. At the moment, these Sync tasks don't allow you to have all your Kestra specific objects have Kestra as the source of truth, and Namespace Files have GIT as the source of truth.

There's a few options I can imagine:
- `sourceOfTruth` is split into multiple properties. Kestra specific source of truth and Namespace source of truth
- Take it a step further and allow you to configure each objects source of truth (but at that point, you might as well use the specific tasks rather than modify this task.... however those only let you target specific namespaces, not entire tenants.

jymaire:
## Summary

- Adds an optional `sourceOfTruthOverrides` property (`flows`, `namespaceFiles`) to `TenantSync` and `NamespaceSync`, so a single run can push one resource kind to Git while pulling the other kind from Git. Resolution is `override.<kind> ?? sourceOfTruth ?? KESTRA`; leaving it unset keeps today's behavior byte-for-byte.
- Direction-gated side effects (namespace discovery/creation, `ensureNamespaceFolders`, the `gitHasContent` probe, diff output selection) are gated per-kind via `anyGit`/`anyKestra` instead of the single `sourceOfTruth` value. A mixed run always emits the in-memory diff (`DiffLine.writeIonFile`) since `createIonDiff` only sees Git working-tree changes and would miss a Git->Kestra pull.
- `TenantSync`/`NamespaceSync`'s nested `SourceOfTruth` enum moves to the shared kernel (`io.kestra.plugin.git.shared.SourceOfTruth`, from `plugin-git-lib` 1.2.0-SNAPSHOT) alongside the new `SourceOfTruthOverrides` type — both genuinely shared with `plugin-ee-git`.
- Fixes a pre-existing bug in `NamespaceSync#deleteNamespaceFile`, uncovered while writing the mixed `whenMissingInSource: DELETE` test: it passed `NamespaceFile#storagePath()` (Kestra-URI-prefixed) to `Namespace#delete`, whose contract expects the same namespace-relative path shape as `Namespace#putFile` — the mismatch silently no-op'd instead of deleting the file.
- Adds a `sourceOfTruth: KESTRA` + `sourceOfTruthOverrides.namespaceFiles: GIT` example to both tasks.

closes: https://github.com/kestra-io/plugin-git/issues/237

## Test plan

- [x] `rtk test ./gradlew test` passes (`CloneTest`'s two self-hosted-Gitea tests fail only for lack of a local Gitea instance in this sandbox — pre-existing, unrelated to this change)
- [x] New tests cover: mixed push+pull in one run, mixed + dryRun (single diff with both directions), mixed + `whenMissingInSource: DELETE` (deletes on the correct side per kind), no-override/degenerate-override byte-identical diffs, and the narrowed `gitHasContent` probe / `anyGit` namespace auto-creation

---
*[View as Artifact](https://claude.ai/artifact/PTyhNvrtD4oEFZCKQNF5zp)*


jymaire:
## QA report — mixed source of truth (`sourceOfTruthOverrides`)

**Edition**: EE | **Docker tag**: `develop` (next: 2.0.1)
**Instance**: http://localhost:8390 (left running for manual review — `docker rm -f kestra-psi7h` to stop)
**Screenshots**: https://claude.ai/artifact/Bt68xTQS2YjFspgWbxYENG

### Result summary

| Flow | Scenario | Status |
|------|----------|--------|
| `nsync-mixed` | **mixed**: flows KESTRA→Git + Namespace Files Git→KESTRA in one run (new feature) | ✅ SUCCESS |
| `nsync-nooverride` | no override, `sourceOfTruth: KESTRA` (non-regression) | ✅ SUCCESS |
| `nsync-mixed-dryrun` | mixed + `dryRun: true` (new feature) | ✅ SUCCESS (see note) |

**Verdict: PASS.** The core acceptance criterion is confirmed: a single `NamespaceSync` run pushed a Kestra-owned flow to Git **and** pulled a Git-owned Namespace File into Kestra, with `sourceOfTruth: KESTRA` + `sourceOfTruthOverrides.namespaceFiles: GIT`.

> Scope note: this PR touches the **OSS** `TenantSync`/`NamespaceSync`. Only `NamespaceSync` was exercised end-to-end; `TenantSync` shares the identical per-kind resolution/gating logic (same `SourceOfTruthOverrides.resolve` + `anyGit`/`anyKestra` code paths). The EE port is the explicitly out-of-scope follow-up.

### Environment adjustments (why they were needed)
- The EE `develop` image ships `plugin-ee-git`, which carries its **own** `io.kestra.plugin.git.NamespaceSync`/`TenantSync` (same package) and shadowed the OSS PR JAR. QA therefore mounted a plugins dir containing **only** the OSS `plugin-git` build (`4.1.1-SNAPSHOT`, consuming `plugin-git-lib:1.2.0-SNAPSHOT` from mavenLocal) so the PR code is authoritative. No JARs were deleted.
- A local `file://` bare git repo was used as the remote (not the self-signed Gitea). Reason: `NamespaceSync`/`TenantSync` extend `AbstractCloningTask`/`AbstractKestraTask` and do **not** call `configureEnvironmentWithSsl`, so `trustedCaPemPath` is a no-op for them (pre-existing, unrelated to this PR) and the Gitea cert is a non-CA self-signed cert. `file://` isolates the test from TLS entirely.

### Flow 1: `nsync-mixed` (✅ SUCCESS) — the feature

```yaml
id: nsync-mixed
namespace: qa.tests
tasks:
  - id: sync
    type: io.kestra.plugin.git.NamespaceSync
    namespace: qa.git
    sourceOfTruth: KESTRA
    sourceOfTruthOverrides:
      namespaceFiles: GIT
    whenMissingInSource: KEEP
    url: file:///tmp/qa-git-237.git
    branch: main
    kestraUrl: http://localhost:8390
    auth:
      username: "********"
      password: "********"
```
</details>

**Setup**: Kestra namespace `qa.git` had flow `kestra-owned-flow`; Git had `qa.git/files/hello-from-git.txt` (and no flows).

**Gantt** ([screenshot](https://claude.ai/artifact/Bt68xTQS2YjFspgWbxYENG#nsync-mixed-gantt))

| Task | Status | Duration |
|------|--------|----------|
| `sync` | SUCCESS | 0.75s |
| **Total** | **SUCCESS** | **0.91s** |

**Logs** ([screenshot](https://claude.ai/artifact/Bt68xTQS2YjFspgWbxYENG#nsync-mixed-logs)): `Now in NamespaceSync for namespace qa.git` → `Start cloning from 'file:///tmp/qa-git-237.git'` → `Checked out branch main`. No errors.

**Outcome verified on both sides after the run:**
- Git bare repo now contains `qa.git/flows/kestra-owned-flow.yaml` → **flow pushed Kestra→Git** ✅
- Kestra namespace `qa.git` files now contains `hello-from-git.txt` (65 bytes) → **file pulled Git→Kestra** ✅

Both directions in a single execution — exactly the feature.

### Flow 2: `nsync-nooverride` (✅ SUCCESS) — non-regression

```yaml
id: nsync-nooverride
namespace: qa.tests
tasks:
  - id: sync
    type: io.kestra.plugin.git.NamespaceSync
    namespace: qa.git
    sourceOfTruth: KESTRA
    whenMissingInSource: KEEP
    url: file:///tmp/qa-git-237.git
    branch: main
    kestraUrl: http://localhost:8390
    auth: { username: "********", password: "********" }
```
</details>

With no override (uniform `KESTRA`), the run logged **`No changes to commit.`** and succeeded — Git and Kestra were already consistent from Flow 1. Confirms the pre-existing single-source-of-truth path is unchanged.

### Flow 3: `nsync-mixed-dryrun` (✅ SUCCESS) — with one observation

Same as Flow 1 plus `dryRun: true`. The task completed SUCCESS and applied nothing (correct).

⚠️ **Observation (non-blocking)**: the task-run `outputs` were empty at runtime — the `diff` ION output was not populated even though a push (flow) and a pull (file) were pending. The unit tests (`mixed + dryRun` case) assert the diff file content at code level, so this is likely a runtime output-surfacing nuance rather than a logic bug, but worth a quick confirm from the author that `NamespaceSync` populates `Output.diff` on a mixed dry run.

### Timeouts
None. All executions completed in under 1s.

jymaire:
## Docs impact assessment

The `/kestra-docs:impact-assessor` skill flagged this PR as **likely** to require documentation updates.

Assessment:

- **docs_required**: likely
- **user_impact**: Adds an optional `sourceOfTruthOverrides` property to `TenantSync`/`NamespaceSync` so a single run can push one resource kind to Git while pulling another from Git. Unset → today's behavior, so existing docs don't become wrong, but the new capability (and its `whenMissingInSource` footgun) is user-facing.
- **affected_audience**: plugin users doing Git sync; enterprise admins (EE follow-up)
- **doc_types**: reference (auto-generated from `@Schema`), concept/how-to update in the Version Control / Git sync guide, plus a `@Schema` description improvement in this repo
- **candidate_paths**: `kestra-io/docs` Version Control / Git sync guide (mixed source-of-truth section) — separate docs-repo follow-up; in this repo, the `sourceOfTruthOverrides` `@Schema` description
- **reasoning**:
  - New user-visible plugin property + new `@Plugin` examples = high docs signal.
  - Reference docs are auto-generated from `@Schema`, so they update automatically — **but** the current `sourceOfTruthOverrides` description does not mention the key footgun: with a mixed direction + `whenMissingInSource: DELETE`, a Namespace File present in Kestra but absent from Git is **deleted from Kestra** (Git wins for files). The plan explicitly called for the descriptions to carry this weight.
- **gaps_or_unknowns**:
  - The narrative Git-sync guide lives in `kestra-io/docs` and is a separate follow-up (not blocking this PR).

**Actionable in this PR**: extend the `sourceOfTruthOverrides` `@Schema` description (and/or the mixed `@Example`) to warn about the per-kind `whenMissingInSource: DELETE` semantics, so the auto-generated reference docs surface the footgun.

@aj-emerich FYI — docs impact flagged (likely) on this PR.

build.gradle
```diff
@@ -55,7 +55,7 @@ dependencies {
     compileOnly group: "io.kestra", name: "script", version: kestraVersion
 
     // shared Git kernel (also brings jgit and kestra-api-client transitively as `api`)
-    api 'io.kestra.plugin:plugin-git-lib:1.1.0'
+    api 'io.kestra.plugin:plugin-git-lib:1.2.0-SNAPSHOT'
```
jymaire: 🟠 [Guidelines] SNAPSHOT dependency on plugin-git-lib
Problem: pins to plugin-git-lib:1.2.0-SNAPSHOT (companion PR kestra-io/plugin-git-lib#19 is still open/unreleased), producing a non-reproducible build if merged as-is.
Fix: merge and release plugin-git-lib 1.2.0 first, then bump this to the released version before merging this PR (same sequencing used for the 1.1.0 release).

src/main/java/io/kestra/plugin/git/NamespaceSync.java
```diff
@@ -285,7 +318,7 @@ public Output run(RunContext runContext) throws Exception {
         syncNamespaces.add(rNamespace);
         if (rIncludeChildNamespaces) {
             syncNamespaces.addAll(descendantNamespaces(runContext, tenantId, rNamespace));
-            if (rSourceOfTruth == SourceOfTruth.GIT) {
+            if (anyGit) {
                 for (String gitNamespace : gitDescendantNamespaces(baseDir, rNamespace)) {
```
jymaire: 🟠 [Correctness] Missing per-kind narrowing on child-namespace auto-creation
Problem: this loop is gated only on anyGit, but gitDescendantNamespaces(baseDir, rNamespace) (defined below at :816) matches any child directory with EITHER flows/ OR files/, with no check against which specific kind (flowsSource/filesSource) is actually GIT-sourced. In mixed mode (e.g. sourceOfTruth: KESTRA, sourceOfTruthOverrides.namespaceFiles: GIT) with includeChildNamespaces: true, a Git-only child namespace whose only content is a flows/ directory gets spuriously auto-created as an empty Kestra namespace, even though flows stay Kestra-sourced in this run. TenantSync fixed exactly this false-positive with gitHasContent(namespaceRoot, flowsSource, filesSource) (TenantSync.java:907) but the equivalent NamespaceSync path was not updated.
Fix: narrow this loop the same way TenantSync.gitHasContent does — only auto-create when the git namespace directory actually has content for a kind that resolved to GIT: (flowsSource==GIT && flows/ exists) || (filesSource==GIT && files/ exists).

src/main/java/io/kestra/plugin/git/NamespaceSync.java
```diff
@@ -266,6 +290,15 @@ public Output run(RunContext runContext) throws Exception {
 
         var rGitDirectory = runContext.render(this.gitDirectory).as(String.class).orElse(null);
         var rSourceOfTruth = runContext.render(this.sourceOfTruth).as(SourceOfTruth.class).orElse(SourceOfTruth.KESTRA);
+        var flowsSource = SourceOfTruthOverrides.resolve(
```
jymaire: 🟡 [Clean Code / DRY] Duplicated per-kind resolution block
Problem: this 6-line derivation of flowsSource/filesSource/anyGit/anyKestra/mixed is duplicated verbatim in TenantSync.run() (TenantSync.java:260-266). NamespaceSync and TenantSync can't share a task superclass, but the derivation itself doesn't need to live at the call site.
Fix: add a resolveAll(RunContext, SourceOfTruthOverrides, SourceOfTruth fallback) helper to SourceOfTruthOverrides in the shared lib returning a small record/holder with flowsSource, filesSource, mixed — both tasks call it once instead of re-deriving.

jymaire (commented):
## Kestra Plugin Code Review

Issue #237 asks for splitting `sourceOfTruth` so Kestra-native objects (flows) and Namespace Files can have independent directions. The PR implements exactly the first option proposed in the issue (`sourceOfTruthOverrides.flows`/`.namespaceFiles`) with `override ?? sourceOfTruth ?? KESTRA` fallback semantics, scoped correctly to this repo's tasks (flows + Namespace Files only — Tests/Dashboards/Apps are explicitly left as an EE extension point via the non-final `SourceOfTruthOverrides` base class). No scope creep.

- 🟠 `build.gradle`: dependency pinned to `plugin-git-lib:1.2.0-SNAPSHOT` — must be bumped to a released version once companion PR kestra-io/plugin-git-lib#19 is merged and released, per this repo's own convention (see prior 1.1.0 release cycle). Merging against a SNAPSHOT produces non-reproducible builds.
- 🟡 DRY: the 6-line `flowsSource`/`filesSource`/`anyGit`/`anyKestra`/`mixed` resolution block is duplicated verbatim in `NamespaceSync.run()` and `TenantSync.run()`. Since these tasks can't share a common ancestor (different base classes: `AbstractCloningTask` vs `AbstractKestraTask`), consider hoisting a single `SourceOfTruthOverrides.resolveAll(RunContext, SourceOfTruthOverrides, SourceOfTruth fallback)` helper into the shared lib that returns a small record (flowsSource, filesSource, mixed) instead of duplicating the derivation at each call site.
- 🟡 Nested enum removal (`NamespaceSync.SourceOfTruth` / `TenantSync.SourceOfTruth` → shared `io.kestra.plugin.git.shared.SourceOfTruth`) is transparent at the YAML/flow level (same enum names GIT/KESTRA) but is a breaking change for any Java code referencing the old nested type directly. AGENTS.md states this plumbing is shared with `plugin-ee-git`; please grep that repo for `NamespaceSync.SourceOfTruth` / `TenantSync.SourceOfTruth` before merge.

No new credential handling, no new external I/O surface, no template-injection surface introduced. The `deleteNamespaceFile` fix (removing the fully-qualified `NamespaceFile#storagePath()` lookup and the `files/` fallback in favor of `Namespace#delete(Path.of(file))`) was verified against `Namespace` interface bytecode: `delete(NamespaceFile)` already delegates to `delete(Path.of(namespaceFile.path()))`, i.e. the same namespace-relative `Path` shape used by `putFile(Path, InputStream)`. The fix makes delete symmetric with put and correctly targets only the resolved `rel` path — no over-delete risk observed, and this is a genuine bug fix (prior code silently no-op'd on delete). Path-traversal exposure on `Path.of(rel)` for put/delete is pre-existing (unchanged by this PR), not introduced here.

### Performance — 0 issues found
No new unbounded allocation, no new loops over remote data, no blocking-in-loop patterns introduced by the per-kind gating changes.

### Correctness — 1 finding
- 🟠 `NamespaceSync.java:322-334` (child-namespace auto-creation from `gitDescendantNamespaces`, def at :816-828): gated only on `anyGit`, but `gitDescendantNamespaces` matches any child directory containing EITHER flows/ OR files/, with no check against which specific kind is actually GIT-sourced. See inline comment for the concrete mixed-mode false-positive scenario and fix.

### QA follow-up — resolved by code reading
Re: the non-blocking QA observation about an empty `Output.diff` on a mixed dry run — both `NamespaceSync` and `TenantSync` compute `diffFile = DiffLine.writeIonFile(runContext, diffs)` unconditionally in the dry-run branch, and `diffs` is populated by unconditional `diff.add(...)` calls in `planFlows`/`planNamespaceFiles` (only `apply.add(...)` is gated by `!rDryRun`). The code path is correct; the empty diff observed in QA was most likely environmental (e.g. no actual cross-direction change existed in that specific run, or an ION-file rendering/download issue in the UI) rather than a defect in this PR. Recommend a targeted re-run with explicit before/after file state to confirm.

**Verdict:** REQUEST CHANGES

src/main/java/io/kestra/plugin/git/TenantSync.java
```diff
@@ -230,6 +257,16 @@ public Output run(RunContext runContext) throws Exception {
         String rGitDirectory = runContext.render(this.gitDirectory).as(String.class).orElse(null);
         SourceOfTruth rSourceOfTruth = runContext.render(this.sourceOfTruth).as(SourceOfTruth.class)
             .orElse(SourceOfTruth.KESTRA);
+        SourceOfTruthOverrides.Resolved resolvedSource = SourceOfTruthOverrides.resolveAll(
```
jymaire: 🟡 [Guidelines] Use var for local variables
Problem: This block declares SourceOfTruthOverrides.Resolved resolvedSource, SourceOfTruth flowsSource, SourceOfTruth filesSource, boolean anyGit, boolean mixed with explicit types; the equivalent NamespaceSync.java block uses var throughout, so the two copy-pasted call sites have drifted stylistically.
Fix: Switch these five locals to var per the Java Style guideline (explicit types only when needed for readability/disambiguation, which isn't the case here).

jymaire (commented):
## Kestra Plugin Code Review — Re-review (cycle 2)

The design (`sourceOfTruthOverrides.flows`/`namespaceFiles`, falling back to `sourceOfTruth`) matches option 1 proposed in kestra-io/plugin-git#237 exactly: Kestra objects keep one source of truth, Namespace Files get their own.

### Cycle-1 findings — verification

1. **Child-namespace auto-creation narrowing (was must-fix) — RESOLVED.** `gitHasContent(namespaceRoot, flowsSource, filesSource)` is now a shared `protected static` method on `AbstractGitTask` (lib), used identically by `NamespaceSync#gitDescendantNamespaces` and `TenantSync`'s child-namespace creation check, replacing the old ungated `flows/ OR files/` probe. Verified the regression test `NamespaceSyncTest#includeChildNamespaces_gitOnlyChildWithOnlyFlows_staysKestraSourced_notTouchedByFileSourceOfTruth`: it seeds a child namespace that is Git-only for flows, keeps `sourceOfTruth: KESTRA` (flows) with `sourceOfTruthOverrides.namespaceFiles: GIT`, and `whenMissingInSource: DELETE`. Traced the old (pre-fix) code path by hand — the child would have matched on `flows/` presence alone and been pulled into `syncNamespaces`, and since `flowsSource == KESTRA`, the Git-only flow (missing from Kestra) would then have been deleted from Git — so the assertion genuinely fails without the fix. `TenantSyncTest#gitHasContent_narrowsProbeToResolvedSourcePerKind` independently unit-tests the shared method via reflection with both a true and a false-positive-guard case. Sound.

2. **SNAPSHOT dependency — ESCALATED to 🔴 Block, no longer a benign merge-checklist item.** `build.gradle` pins `plugin-git-lib:1.2.0-SNAPSHOT`, but the actual merged state of that dependency (PR kestra-io/plugin-git-lib#19, squash-merged as `f673df5` on lib `main`) contains only `SourceOfTruth` + the base `SourceOfTruthOverrides` (with just `resolve`) — **not** `SourceOfTruthOverrides.resolveAll`/`Resolved`, nor `AbstractGitTask.gitHasContent`. Those live only in an unpublished, unreviewed commit `31cc26b` ("feat(shared): add resolveAll + shared gitHasContent for per-kind source of truth") sitting on the lib's `feat/237-source-of-truth-overrides` branch with **no open PR** (`gh pr list` on plugin-git-lib shows nothing beyond merged #19). Confirmed empirically: PR #357's `check / main` CI run is currently **failing at `compileJava`** with `cannot find symbol` for exactly `SourceOfTruthOverrides.resolveAll` and `gitHasContent` in both `NamespaceSync.java` and `TenantSync.java` (job 106358686218). The PR only builds locally because the contributor's machine has `31cc26b` published to `~/.m2` via `publishToMavenLocal`; no other machine, including CI, can build this branch as-is. This must be fixed (open a lib PR for `31cc26b`, get it merged and published, or fold it back into a re-opened #19) before merge — it is not the previously-accepted "deferred Central release" gap, it is a broken build.

3. **DRY (`SourceOfTruthOverrides.resolveAll`) — RESOLVED.** Both tasks now call the shared `resolveAll`; `anyGit`/`anyKestra`/`mixed` derivation (`flows == GIT || files == GIT`, `flows == KESTRA || files == KESTRA`, `flows != files`) is unchanged in semantics from the pre-refactor inline logic. The mixed-mode diff-source selection (`mixed || flowsSource == GIT` picking the in-memory `diffs` list over `createIonDiff`) is correctly derived in both `NamespaceSync` and `TenantSync`, and matches the stated rationale (a Kestra→Git flow push + Git→Kestra file pull only has the push reflected in git-status).

4. **Nested `SourceOfTruth` enum removal / EE compat — unchanged since cycle 1, no new information contradicts the earlier reasoning.** Carried forward as sound.

Previously-cleared items re-confirmed: `deleteNamespaceFile` now calls `runContext.storage().namespace(namespace).delete(Path.of(file))` (namespace-relative shape, matching `putFile`), with a comment explaining the earlier silent no-op. Mixed dry-run `Output.diff` is populated via `DiffLine.writeIonFile` in both tasks per point 3.

- 🟡 [Guidelines] Inconsistent `var` usage — `TenantSync.java:260-266` declares `SourceOfTruthOverrides.Resolved resolvedSource`, `SourceOfTruth flowsSource`, `SourceOfTruth filesSource`, `boolean anyGit`, `boolean mixed` with explicit types, while the equivalent block in `NamespaceSync.java` correctly uses `var` throughout. Per the Java Style guideline ("Use `var` for all local variables whenever it is legally possible"), none of these need an explicit type for readability. Non-blocking, but the two call sites should stay in lockstep since they're copy-pasted logic.

### Security (OWASP Top 10:2025 + KPS) — no issues found
No new HTTP calls, secrets, deserialization, or template rendering paths introduced; `sourceOfTruthOverrides` is a plain enum-valued config resolved through `runContext.render`, consistent with the rest of the file.

### Performance — no issues found
`gitHasContent` is O(1) (two directory-existence checks); no new loops, allocations, or blocking I/O added beyond what already existed.

**Verdict:** BLOCK — the branch does not compile in CI because `plugin-git-lib` PR #19 was merged without the `resolveAll`/`gitHasContent` commit (`31cc26b`) that `NamespaceSync`/`TenantSync` depend on. Once that lib commit has its own reviewed/merged PR and a real `1.2.0-SNAPSHOT` (or released) artifact containing it is available to CI, this is otherwise ready — only a 🟡 nit remains.

*(Posted as a COMMENT review, not REQUEST_CHANGES, because GitHub does not allow the PR author to request changes on their own pull request via the API — treat the verdict below as REQUEST_CHANGES/BLOCK.)*

jymaire:
### Re: cycle-2 BLOCK — build-integrity resolved

The BLOCK correctly caught that the `resolveAll`/`Resolved`/`AbstractGitTask.gitHasContent` code this PR calls was stranded: lib PR #19 was squash-merged with only its first commit, leaving the refactor commit on a branch with no open PR (and that branch was also behind `main`, so it couldn't simply be re-PR'd).

Resolved:
- The stranded shared code is now in a clean lib PR off current `main`: **kestra-io/plugin-git-lib#20** (`resolveAll` + shared `gitHasContent`), cherry-picked so it does not touch unrelated files.
- 🟡 `var` nit fixed — `TenantSync` resolved-source locals now use `var` to match `NamespaceSync` (`1bcae5b`).

Cycle-2's other items were already verified resolved/sound (#1 child-namespace narrowing + regression test, #3 `resolveAll`, #4 EE enum compat, and the `deleteNamespaceFile` fix).

**Remaining pre-merge gate (unchanged, maintainer-owned):** this PR pins `plugin-git-lib:1.2.0-SNAPSHOT`. `check / main` stays red until `plugin-git-lib` `1.2.0` — including #19 (merged) **and** #20 — is released so CI can resolve it; then this PR's dependency is bumped `1.2.0-SNAPSHOT` → `1.2.0`. No code changes remain.

Malaydewangan09:
@jymaire we need to do the same in EE git also?

jymaire:
> @jymaire we need to do the same in EE git also?

It doesn't seem to be used in Kestra EE. But as `sourceOfTruth` is declared in plugin-git-lib, it might be interessant to move my fix into the lib. I check it, thanks

jymaire:
> @jymaire we need to do the same in EE git also?

@Malaydewangan09 after investigation: in OSS TenantSync, NamespaceSync can sync multiple flow and namespace file that was the point of sourceOfTruthOverrides.

In EE, tasks are only of a type and there is no reference too sourceOfTruth so my conclusion is no need to report to EE nor lib.

src/main/java/io/kestra/plugin/git/NamespaceSync.java
jymaire: Resolved in 7de1357 — child-namespace auto-creation now filters on the shared `gitHasContent(p, flowsSource, filesSource)` (NamespaceSync.java:829), so a Git-only child holding only a `flows/` directory is no longer spuriously auto-created when only files resolve to GIT. Covered by `NamespaceSyncTest#includeChildNamespaces_gitOnlyChildWithOnlyFlows_staysKestraSourced_notTouchedByFileSourceOfTruth`.

src/main/java/io/kestra/plugin/git/NamespaceSync.java
jymaire: Resolved — the per-kind derivation is hoisted into `SourceOfTruthOverrides.resolveAll(...)` in the shared lib. Both `NamespaceSync.run()` (:293) and `TenantSync.run()` (:260) now call it and read `flows()`/`namespaceFiles()`/`anyGit()`/`mixed()` off the returned `Resolved` record instead of re-deriving the block inline.

src/main/java/io/kestra/plugin/git/TenantSync.java
jymaire: Fixed in 1bcae5b — the resolved-source locals in `TenantSync` now use `var` throughout, matching the `NamespaceSync` block.

build.gradle
jymaire: Leaving this open as the maintainer-owned pre-merge gate. The pin stays `plugin-git-lib:1.2.0-SNAPSHOT` until lib `1.2.0` (including #19 + #20) is released; only then does this line get bumped to the released `1.2.0` and `check / main` can go green. No code change until then, so this thread stays open to track the blocker.

jymaire:
Following up on my earlier answer — after digging into the actual code I need to correct it: EE **does** need this, and the lib was already involved.

- EE ships its own `TenantSync`/`NamespaceSync` in package `io.kestra.plugin.git` (they shadow the OSS classes at runtime). They're supersets of OSS — a single `sourceOfTruth` drives flows + namespace files **plus** apps, unit tests, blueprints, dashboards. So `sourceOfTruthOverrides` applies there too, and EE users would otherwise never get it. The single-kind EE tasks (`SyncApps`, `SyncBlueprints`, `SyncDashboards`, `SyncUnitTests`) are one-kind-each and correctly need nothing.
- Lib: this PR already moves `SourceOfTruth` + `SourceOfTruthOverrides` into `plugin-git-lib` (now released as **1.2.0** on Maven Central). The shared `SourceOfTruthOverrides` was made non-final specifically so EE can extend it with its own kinds.

Done: EE port implemented and QA'd — kestra-io/plugin-ee-git#179 (extends the shared overrides with apps/unitTests/blueprints/dashboards, consumes lib 1.2.0). Verified end-to-end on Kestra EE `develop`: a mixed `sourceOfTruth: KESTRA` + `sourceOfTruthOverrides.namespaceFiles: GIT` run pushed flows Kestra→Git and pulled namespace files Git→Kestra in a single execution.

This PR's `build.gradle` is now pinned to the released `plugin-git-lib:1.2.0` (was `-SNAPSHOT`).

Malaydewangan09:
Reviewed at `644b988`. The shape is right, and notably this is the reference implementation the EE port should follow: both tasks call the shared `SourceOfTruthOverrides.resolveAll(...)` (`TenantSync.java:260`, `NamespaceSync.java:293`) and the inherited `AbstractGitTask.gitHasContent(...)` (`TenantSync.java:310`, `NamespaceSync.java:829`), with no local reimplementation. plugin-ee-git#179 hand-rolls both in each task instead, I will point it here.

One thing to fix before merge.

### `protectedNamespaces` does not guard every delete in `TenantSync`

`TenantSync.java:646-649`, in `planNamespaceFiles`, the file-exists-only-in-Git plus files-source-KESTRA plus `DELETE` branch:

```java
case DELETE -> {
    diffs.add(DiffLine.deletedGit(filePath.toString(), rel, Kind.FILE));
    if (!rDryRun)
        apply.add(() -> deleteGitFile(filePath));
}
```

No `isProtected` check. The other three delete branches in the same file have one: flows-from-git at `:530`, flows-from-Kestra at `:554`, files-from-Kestra at `:668`. `NamespaceSync` guards both of its delete-from-git branches too, `:455` for flows and `:558` for files. This single branch is the odd one out across both tasks.

So with `sourceOfTruth: KESTRA`, `protectedNamespaces: [system]` and `whenMissingInSource: DELETE`, a file under `system/files/` in Git that Kestra does not have gets deleted from Git and pushed, despite `system` being protected.

Pre-existing on `main`, and this PR only renames `rSourceOfTruth` to `filesSource` there. But this is the PR unifying the per-kind delete semantics and selling a safety story around `protectedNamespaces`, so it is the natural place to close it rather than carry the inconsistency into a feature whose whole point is more complex delete behaviour.

### Test gaps

- Nothing exercises `protectedNamespaces` together with mixed sources and `DELETE`. Such a test would have caught the above.
- `TenantSync` has no end-to-end mixed run. `NamespaceSyncTest` gained four real `task.run(rc)` tests (push and pull in one run, dry-run with both directions in one diff, `DELETE` on both sides, unset versus degenerate-override equality). `TenantSyncTest`'s new tests all go through `getDeclaredMethod` reflection on `planFlows`/`planNamespaceFiles`, so nothing proves the top-level `resolveAll` wiring, the diff-file selection branch, or the auto-create gating in `TenantSync.run()`. This mirrors the pre-existing asymmetry, `TenantSyncTest` does not extend `AbstractGitTest`, so it is not a corner this PR cut. Still a gap for the amount of new branching added to that method.

### Verified good

- The `deleteNamespaceFile` fix is correct against core's contract, `Namespace.delete(NamespaceFile)` delegates to `delete(Path.of(file.path()))`, the namespace-relative shape. Worth recording why the removed two-attempt fallback was safe to drop: `InternalNamespace.delete(Path)` looks the path up against metadata keyed on that relative shape and returns an empty list rather than throwing, so the old `catch (IOException)` could never fire and the `files/` retry was dead code. That also means `DELETE` with `sourceOfTruth: GIT` silently never deleted stale namespace files on `main` today, which is worth a line in the release notes.
- Unset `sourceOfTruthOverrides` really does collapse every new branch back to the old behaviour, since `resolve` returns the fallback when the override is null.
- The `mixed || flowsSource == GIT` diff selection is deliberate and correct, not an oversight. `createIonDiff` uses `git.diff().setCached(true)`, an index-versus-HEAD diff, so in a mixed run it would miss the kind that only makes Kestra API calls and touches no local file. Falling back to the manually tracked `diffs` list whenever `mixed` holds is the right call.
- `shouldIgnoreGlobalDashboardsDirectoryWhenDiscoveringNamespaces` covers the `_global` discovery edge already.

### Minor

`src/main/resources/doc/io.kestra.plugin.git.md` is unchanged. It does not mention `TenantSync` or `NamespaceSync` at all today, so this is a pre-existing gap rather than something this PR introduced, but `sourceOfTruthOverrides` is a footgun-adjacent property that got a dedicated `@Schema` writing pass, and it appears nowhere in the plugin overview.

Malaydewangan09 (approved):

